Issue #268, opened 2014-01-24
Start processes as different users (security)


achingbrain:
If I've got a few node processes running under pm2, for security purposes I don't necessarily want them to run as the same user.

Worryingly, it seems that if I start pm2 as root, then switch to user foo to start a process, the process runs as root and not foo, which seems like a privilege escalation vulnerability.

Also, the process logs end up in foo's home directory but owned by root which seems like a side effect of the above.

PR #329, opened 2014-03-03
Adds option to switch to a different user/group before starting a managed process

achingbrain:
This is an attempt to address part of https://github.com/Unitech/pm2/issues/268 - if you run pm2 as root, everything it starts also runs as root.

It adds two options to the command line `--run-as-user` and `--run-as-group`.

If you are running `pm2` as root, and you start a process like this:

```
# pm2 --run-as-user foo start test.js
```

..then pm2 will fork the current process (as `root`), set up the child's logs and pid file (as `root`, in /root/.pm2), then switch to user `foo` to require and run `test.js`.  Because the logs were opened before the privilege drop the process can still write to them even though the new user doesn't have the permissions to do so.

If you then do `ps aux | grep pm2` you'll see something like:

```
root            31568   0.0  0.0  2432784    476 s009  R+    7:30PM   0:00.00 grep pm2
foo             31564   0.0  0.1  3057912  23156   ??  S     7:30PM   0:00.10 pm2: test 
root            31561   0.0  0.1  3053776  19968   ??  Ss    7:30PM   0:00.12 pm2: Satan Daemonizer 
```

..which shows that `pm2` is still running as `root` whereas the managed process `test.js` is running as `foo`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,10 @@ $ pm2 start app.json                # Start processes with options declared in a
 $ pm2 start app.js -i max -- -a 23  # Pass arguments after -- to app.js
 
 $ pm2 start app.js -i max -e err.log -o out.log  # Will start and generate a configuration file
+
+$ pm2 --run-as-user foo start app.js  # Start app.js as user foo instead of root (pm2 must be running as root)
+
+$ pm2 --run-as-user foo --run-as-group bar start app.js  # Start app.js as foo:bar instead of root:root (pm2 must be running as root)
 ```
 
 You can also execute app in other languages ([the fork mode](#a23)):


aspnetde:
Just a thought: I found PM2 absolutely useful and well designed, it seemed to be the solution I was looking for. I really appreciate the work that is done here, but not implementing and ignoring such a fundamentally important feature as privilege separation is unfortunately a show-stopper.

You're right: if using only one Node application/website per server, the current setup is sufficient. But as soon as I have at least two sites running side by side, I don't want to run them with the same account, risking to affect each other in case of any vulnerability. I am not a bad developer, but who can say their code is 100% secure? I am not that guy ;-).

So I need to separate all of the "instances". I hope you'll find the time and motivation to implement that feature in the future.

Unitech:
Yes it's done!

With UNIX socket feature you can now run totally separated instances of PM2.

You can either overidde the PM2_HOME variable to create another instance of PM2 for the same user:

```bash
$ PM2_HOME='/home/myuser/.pm2' pm2 start app1.js --name="app-one"
$ PM2_HOME='/home/myuser/.pm2' pm2 list 

$ PM2_HOME='/home/myuser/.pm3' pm2 start app2.js --name="app-two"
$ PM2_HOME='/home/myuser/.pm3' pm2 list
```

Or under different users:
```bash
user-1$ pm2 start app1.js

user-2$ pm2 start app2.js
```

In this case the user-1 will not be able to list the process of the pm2 started under user-2 and vice versa

dandv:
Does this work as of May 2015? See also https://github.com/Unitech/PM2/issues/992#issuecomment-72733760

I'm trying to [start Ghost as user `ghost`](https://allaboutghost.com/keep-ghost-running-with-pm2/) but to create a startup script after `pm2 dump`, I need to run pm2 as root. When I run it from a root login, it doesn't see any pm2 processes.

soyuka:
@adamscybot For this to work we'd need to connect to every RPC available on the server. I think it's possible but it would add a lot of complexity (see Interactor).

aaronroberson:
First off, I love PM2, you guys are killing it!

The comment from @adamscybot is spot on regarding the need to have root behave as an overlord of all the processes. 

Currently if I run any utility commands (pm2 list, monit, etc.) as the root user it spawns a new process which makes a small footprint in memory. 

I would love to see it PM2 updated so that the root user can list, monit and perform other utility operations with access to all processes for all users and without spawning a new process of it's own.

slavafomin:
Hello!

Thank you for this great module!

I've setup PM2 to automatically start with the system under the `pm2` user, however, I want to start each application under it's own unix user, otherwise it would be a high security risk. I've already created all the users and configured ACL entries on the filesystem properly. Is it possible to do this with PM2?

All my apps are declared in the single `apps.json` file, e.g.:

```json
{
	"apps": [
		{
			"name": "foo",
			"cwd": "/var/apps/foo/current",
			"script": "app.js",
			"env": {
				"NODE_ENV": "production",
				"PORT": 8001
			}
		}
	]
}
```

And I would prefer to control all the apps from the single `pm2` user if possible. Any suggestions will be highly appreciated. Thank you!

EmixMaxime:
Hello, I've the same problem as @slavafomin. Any news ? It has been a long time... 

Thank you for your understanding and pm2 !

icharge:
Any update? I'm looking for PM2 startup with user.  This still startup as root.

knoxcard:
Best solution for me, added these two lines to the top of my app.js file...

```
process.setgid('www-data');
process.setuid('www-data')
```

Issue #308, opened 2014-02-16
Systemd init script support


benkaiser:
Related to issue #55 
With the recent news of both [Debian](http://www.phoronix.com/scan.php?page=news_item&px=MTYwMDQ) and [Ubuntu](http://www.phoronix.com/scan.php?page=news_item&px=MTYwNDE) moving to systemd, it makes sense that `pm2 startup` should add support for systemd startup scripts, perhaps via `pm2 startup systemd`. This would also bring support for Arch Linux and Fedora that already use systemd.

PR #333, opened 2014-03-06
Add startup support for systemd based distros

willianpts:
With this commit, systemd can stop, restart and reload pm2.
Tested on Archlinux and Fedora 20.

Relates to #308


grahamrhay:
shouldn't the unit have a value for Restart?

http://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=

PR #459, opened 2014-05-13
Stripping environment variables from the HttpInterface

MichielDeMey:
This gives the user to expose or strip environment variables from the HttpInterface.
By default, they are not exposed due to security reasons.

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -33,6 +33,7 @@ var default_conf = {
 
   DEBUG              : process.env.PM2_DEBUG || false,
   WEB_INTERFACE      : parseInt(process.env.PM2_API_PORT)  || 9615,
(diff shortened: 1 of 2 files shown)


jrottenberg:
This is a pretty nice addition to pm2, but the wording of the issue is misleading.

With that patch the default will be to hide all the environment variables on the web interface, user can get them back with an env flag (PM2_WEB_STRIP_ENV_VARS).

MichielDeMey:
Yes, this PR strips all environment variables by default unless `PM2_WEB_STRIP_ENV_VARS` is set to `false`.

This is because we expose keys/passwords through environment variables rather than hardcoding them.

LukasBeaton:
Why hasn't this been merged in yet?

Any plans to merge this in?

vmarchaud:
@LukasBeaton This one has conflict with development, feel free to re-open one up to date with these changes :)

PR #665, opened 2014-08-29
Better Programmatic API

rlidwka:
With this patch you can do something like:

```js
var pm2 = require('pm2')()
   .on('error', console.error)
   .on('log', console.log)
   .on('connect', function() {
      console.log('connected to pm2')
   })
   .on('disconnect', function() {
      console.log('disconnected')
   })

pm2.app('test')
   .script('/tmp/test.js')
   .instances(4)
   .error_file('/tmp/test-err.log')
   .out_file('/tmp/test-out.log')
   .on('stdout', console.log)
   .on('stderr', console.error)
   .start()

pm2.app('test2')
   .script('/tmp/test2.js')
   .instances(1)
   .start()

setTimeout(function() {
   pm2.list(function(err, list) {
      list.map(function(process) {
         console.log('app %s, pid %s', process.name, process.pid)
      })
   })
}, 2000)
```

1. You don't need to start `pm2.connect()`. It can connect and disconnect automatically.
2. `pm2` is turned into class, which means in the future you can add a feature to connect to multiple pm2 servers (i.e. `new PM2(server1), new PM2(server2)`)
3. `pm2` is turned into eventemitter, which means you can listen on its logs as events for example. This doesn't really work, but it could be.

I just realized that it's very far from be ready to be pulled in (that wrapper around options is awful for example). I could improve that of course, but that's a huge work, and I'm not completely sure it'll be appreciated.

But if you try to improve programmatic api further, here are a few nice ideas to consider.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,1 +1,1 @@
-module.exports = require('./lib/CLI.js');
+module.exports = require('./lib/PM2.js')
--- a/lib/PM2.js
+++ b/lib/PM2.js
@@ -1,0 +1,106 @@
+// public api wrapper
+
+var util = require('util')
+var EE = require('events').EventEmitter
+var CLI = require('./CLI')
+var Common = require('./Common')
+
+module.exports = PM2
+module.exports.PM2 = PM2
+module.exports.PM2App = PM2App
+
+function PM2() {
+  if (!(this instanceof PM2)) return new PM2();
+  this._connected = false;
+  this._connecting = false;
+  this._actions = 0;
+}
+
+util.inherits(PM2, EE);
+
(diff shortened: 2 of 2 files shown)


AnneTheAgile:
Is this PR going to be accepted @Unitech ? It seems like you liked it last year.
AnneTheAgile , CodeTriage fan

rlidwka:
@AnneTheAgile 

It's a stub, not a complete PR. I continued the work in [api](https://github.com/Unitech/PM2/tree/api) branch, which didn't go as well because of technical reasons.

I think I'll close it to avoid bloating open PR list for now.

Issue #304, opened 2014-02-14
gracefulStop ?


ghost:
Hi. I was wondering if there is a way to stop a process gracefully, without reloading it?

Unitech:
#1286

PR #750, opened 2014-10-01
Avoid restart of killed process on "pm2 stop id"

angelix:
If a process for some reason delays to exit, pm2 restarts the process.

This is partial fix commit, fixing only the stop procedure. 
If you want to restart the process, the process is going to spawn two times in a row. One from the restart procedure and one from the monitor procedure.

Example testcase:
try to restart the process and then pm2 list
```javascript
process.on('SIGTERM', function() {
  console.log('Got SIGTERM signal.');
  exit();
});
process.on('SIGINT', function() {
  console.log('SIGINT signal');
  exit();
});
function exit(){
	console.log("Delay exit in 2 secs");
	setTimeout(function(){
		process.exit();
	}, 2000);
}
(function wait () {
    setTimeout(wait, 1000);
})();
console.log("Now you can exit.");
```

### DIFF
--- a/lib/God.js
+++ b/lib/God.js
@@ -76,8 +76,7 @@ God.handleExit = function handleExit(clu, exit_code) {
 
   if (proc.pm2_env.axm_actions) proc.pm2_env.axm_actions = [];
 
-  if (proc.pm2_env.status != cst.ERRORED_STATUS &&
-      proc.pm2_env.status != cst.STOPPING_STATUS)
+  if (proc.pm2_env.status != cst.ERRORED_STATUS)
     proc.pm2_env.status = cst.STOPPED_STATUS;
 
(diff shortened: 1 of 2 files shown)


btleffler:
Is there an issue open for this? It seems like pm2 should wait some configurable amount of time for a process to stop when `pm2 stop <process>` is invoked, and not restart the process immediately. Especially if it's sending `SIGTERM` to the process.

The closest I can find is https://github.com/Unitech/PM2/issues/304 where @Unitech mentions a gracefulStop.

gcochard:
:+1: seen this before.

jshkurti:
https://github.com/Unitech/PM2/blob/master/lib/God.js#L413

PR #790, opened 2014-10-30
fixed `--env` args missing bug

chemdemo:
By default `app.env` is set `(String)production` if starting app with `pm2 start app.js --env pruduction`, but [this line](https://github.com/Unitech/PM2/blob/master/lib/Common.js#L52) `util._extend(app.env, env)` will ignore the `env` argument. So the `process.env.NODE_ENV` will be `undefined` in app, i think it should check the type of `app.env` before extending.

If printing out `process.env.env` in app, it shows `(String)[object Object]`, there is no need to keep it on `process.env` object.


Unitech:
fixed long time ago, thanks anyway for this PR

PR #836, opened 2014-11-22
Fix the wrong base path bug when there is cwd in JSON app declaration

treblam:
When there is a cwd key in JSON app declaration, the script file path should be relative to the cwd key in the JSON file rather than process.cwd(). and so as the watch directories.


szarouski:
Hi guys, do you think it will be useful to be able to use environment variables in cwd?
Use case - I'm lunching `pm2 start` from random locations and type path to json config. If path is relative, then I'll end up with wrong cwd. I could have absolute path, but then I would have to duplicate config for different OS. So something like:
```json
"cwd": "{$PATH_TO_PROJECTS_FOLDER}/path_to_current_project"
```

What do you think?

PR #850, opened 2014-11-28
Configurable CONCURRENT_ACTIONS

kaaelhaa:
In our environment we have made the constant `CONCURRENT_ACTIONS` a configuration variable as this can significantly improve reload times when you have 8 processes of the same app.

The default behaviour is, and should be, to only do one concurrent action. But for power users and environments where it make sense (e.g. can tolerate the overhead of extra processes running while old workers are finishing their jobs in gracefulReload-scenarios) this should be configurable.


kaaelhaa:
So, @Unitech is this PR something you will consider merging (after rebase) or should I close this and continue using our custom fork of PM2?

jorge-d:
Hi @kaaelhaa, this looks OK to me, could you handle the rebase ?

Thanks,

Unitech:
This could hurt some things of the PM2 core and remove rolling restarts, I don't recommend overriding this value

hypesystem:
"could"?

kaaelhaa:
Rebased.

@Unitech From our usage and inspection of the code base I cannot see why this should hurt anything.

It is a power user feature and should be treated as such by having a decent default value. If a user has set `-i max` when starting a process, the underlying hardware has to be capable of handling the extra context-switching and workload for a short period of time (until old processes has finished).

jorge-d:
@jshkurti Any thoughts on this?

jshkurti:
I'm going to refactor a bit reload/gracefulReload soon.
I will keep this PR in mind while doing that and see what is the best solution.
(Chances are high this will be merged unless heavy stability issues but then, user still have the choice to leave the variable at its default value of 1)
I'll keep you informed ;)

jshkurti:
I'm merging this because hey, why the heck not, if you're willing to take the risk, you should be able to do so and keep using the latests versions of PM2 :)
But I won't talk about it in the README because I tried with a value of 4 and some tests didn't pass.

Issue #896, opened 2014-12-22
Can no longer pass command-line args to app


BryanDonovan:
Hi, 

We used to have configs like this:

```
[
    {
        "name": "foo",
        "script": "./app.js",
        "instances": "2",
        "args": "['-e', 'dev']"
    }
]
```

This would make sure pm2 executed the app.js file with the "-e dev" options, similar to starting the app like this:

```
node app.js -e dev
```

This is broken since commit 0c31dede26c214de89a873f9b01da712a8f7c38a.  I didn't see anything in the docs about any changes to this functionality, but perhaps I just missed something.  Is there a new way to do this?

Thanks,
Bryan

PR #898, opened 2014-12-23
fixed critical bug: `process.env` flattens all env-vars.

Tjatse:
Cluster clients can not receive deep-level objects or arrays in the forked process, e.g.:

``` javascript
{
  "args": [
    "foo",
    "bar"
  ], 
  "env": { 
    "foo1": "bar1" 
  }
} 
```

will be parsed to

``` javascript
{
  "args": "foo, bar", 
  "env": "[object Object]"
}
```

Actually it's a `process.env` problem.
#896


rangercyh:
what about forkmode, it just work wrong as before

PR #938, opened 2015-01-13
Never start a process that already has a PID

kanongil:
This fixes an issue where 2 simultaneous `restart`'s on a `fork_mode` process results in 2 running processes.


jshkurti:
Thanks,

Did this happen to you ? (2 running processes)

kanongil:
@jshkurti It has indeed happened to me. I have had issues with errant processes for a long time, and finally decided to do something about it.

PR #982, opened 2015-01-31
Fix for PM2 in Windows when PM2_HOME environment var not set

echiu64:
Easy one line fix...
Optionally ignore the IDEA project files...


echiu64:
Looked at the travis CI logs, don't think the failures are related to my change...

jshkurti:
Okay thanks :)

PR #986, opened 2015-02-02
Windoze

echiu64:
Windows doesn't support SIGQUIT, so running "pm2 kill" doesn't clean up Satan. Using SIGTERM instead.
Added some additional debugging for Windows related issues when it comes to finding PM2_HOME as well.


jshkurti:
looks fine to me, thanks again :)

Issue #993, opened 2015-02-04
arguments sanity check (especially for -i, --instances)


simonjosefsson:
Hi, I played around a bit with pm2 and missed setting a value for number of instances

At least on WIndows, running `pm2 start app.js -i --name "app"` is suicide (**only run this in a VM!**)

![pm2](https://cloud.githubusercontent.com/assets/3242316/6037077/75698500-ac4f-11e4-885c-02cfa73d2c1a.png)

I tried closing the terminals but the whole point of pm2 is to start them again... facepalm. Had to hard-reboot my machine when it became unresponsive.

Is there a usecase for running more processes than cores? If not maybe put a cap at that number or like 2x number of cores. At least something that's less than infinity =)

soyuka:
Did you take this fix into consideration?
https://github.com/Unitech/PM2/issues/584#issuecomment-72720060

simonjosefsson:
No, I installed 0.12.4 from npm. Will try it again with that patch

simonjosefsson:
@echiu64's patch did not affect this (but seems to solve terminals popping up and closing).
See my PR #994 for a suggested fix

soyuka:
:+1:

PR #994, opened 2015-02-04
Fixed improper '--instances' parsing

simonjosefsson:
In #993 I described what happens if you leave out the argument for `--instances`

I've added a simple fix to check if conf.instances is NaN to avoid this


Unitech:
Could you add instances : 1 as an option when doing CLI.start (https://github.com/Unitech/PM2/blob/master/lib/CLI.js#L603) ? I think it's a way to fix this bug

zxqfox:
:+1: for 1

PR #1005, opened 2015-02-10
node don't in  /use/local/bin

rekey:


soyuka:
Update your `$PATH` instead ?

rekey:
there's many versions of node on our server.
so I can't do this.

Issue #1021, opened 2015-02-16
merge_logs flag is causing each process per request when cluster_mode is on to write to the log


askhogan:
Logger is writing multiple logs when merge_logs flag is set to true and cluster_mode is true.

This makes it appear as though all processes are working on the same request, so it makes it appear that the cluster_mode has a bug in it.   However, if the merge_logs flag is off it doesn't occur.  So this is a logging issue.  

SUGGESTION: If using cluster mode and this is occurring to you, set merge_logs to false

process.json

{
  "apps" : [{
    "exec_interpreter"   : "node",
    "exec_mode"          : "cluster_mode",
    "instances"          : 0,
    "log_date_format"    : "YYYY-MM-DD HH:mm Z",
    "max_memory_restart" : "160",
    "merge_logs"         : true,
    "name"               : "testapp",
    "script"             : "server.js",
    "cwd"                : "dashboard",
    "node_args"          : "--harmony",
    "ignoreWatch"        : ["[\/\\]\./", "log"],
    "watch"              : true
  }]
}
dashboard/server.js

var express = require('express');
var app = express();

app.get('/', function (req, res) {
    console.log("req", req.method, req.url);
    res.send('hey from pm2 worker #' + process.env.pm_id + '\n' + new Date());
});

var server = app.listen(3045, function () {
    var host = server.address().address;
    var port = server.address().port;
    console.log('Example app listening at http://%s:%s', host, port);
});
node 0.12 pm2 0.12.5

Run the command pm2 start process.json 
Run the command pm2 logs

Then run a GET command to http://localhost:3045/

pm2 logs outputs a console.log for each worker process

testapp-3 (out): 2015-02-15 14:55 -06:00: req GET /?
testapp-4 (out): 2015-02-15 14:55 -06:00: req GET /?
testapp-2 (out): 2015-02-15 14:55 -06:00: req GET /?
testapp-1 (out): 2015-02-15 14:55 -06:00: req GET /?
testapp-0 (out): 2015-02-15 14:55 -06:00: req GET /?
testapp-7 (out): 2015-02-15 14:55 -06:00: req GET /?
testapp-5 (out): 2015-02-15 14:55 -06:00: req GET /?
testapp-6 (out): 2015-02-15 14:55 -06:00: req GET /?
However in the GET request response you get just one reply from one worker

hey from pm2 worker #6
Sun Feb 15 2015 14:55:01 GMT-0600 (CST)

bwin:
[...]

PR #1022, opened 2015-02-16
Change merge logs due to bug

askhogan:
Merge logs has a bug in it with clsuter mode right now

https://github.com/Unitech/PM2/issues/1021

Change default to not confuse newcomers


soyuka:
#1042

PR #1045, opened 2015-02-26
Fix #1036

### LINKED ISSUE #1036: pm2 now kills detached processes (as of 0.12.2)
I have pm2 managing a node daemon that spawns worker processes as detached children.  The idea was that if the daemon was killed or restarted (after an upgrade, for example) the workers would continue working and publishing to the pub/sub system they work with.  After upgrading to the latest version from 0.10.x, I noticed that pm2 was killing my detached workers as well as the daemon it was managing. 

It looks to me like this is directly related to the addition of the new TreeKill system (https://github.com/Unitech/PM2/commit/fe92e71f7d3629fcea4e804fcfd9cec41e5fb76d) in 0.12.2.  See lines 69-70 of TreeKill:
``` javascript
  var args = downgradePs ? '-eo pid,ppid | grep -w ' : '-o pid --no-headers --ppid ';
  var ps = exec('ps ' + args + ppid, function(err, stdout, stderr){...}
```
That ps command will return the list of all children with that ppid, which, unfortunately, includes children that are detached so that they can live beyond the death of the parent that spawned them.  Rather than sending a signal to the parent and relying on it to clean up the children, it looks like pm2 is now sending `SIGTERM` to all processes in the tree.

Would there be support for an option or config to have pm2 kill processes in the older manner as opposed to this new(er) KillTree method?  It seems to me like there'd be support to kill the parent and have it clean up its children as it sees fit rather than use the nuclear option.

soyuka:
See code, commits and #1036 discussion.

Explanation:

@gcochard said:
> Maybe refer to the kill(2) docs and/or say "negative means group..." rather than "-" here? At first glance this looks like a bug.

From the kill(2) man: 
> If pid is less than -1, then sig is sent to every process in the process group whose ID is -pid.

A child process (eg: `spawn`) is in the same group. Detached ones are not having the same TTY but the same parent id : 

*Node is launching 2 child processes, one detached the other not*

So, we have to SIGINT instead of SIGTERM to make sure that the group only contains processes with the same TTY.

References:
[POSIX signals on wikipedia](http://en.wikipedia.org/wiki/Unix_signal#POSIX_signals)
[Stackexchange on the topic](http://unix.stackexchange.com/a/45435/66995)


lib/God/Methods.js
```diff
@@ -161,6 +161,25 @@ module.exports = function(God) {
...
+   * @method findByPID
+   * @param int pid
+   * @return object|null
+   */
+  God.findOneByPID = function(pid) {
+    var db = God.clusters_db;
+    var arr = [];
+
+    for (var key in db) {
+      if (db[key].pid && db[key].pid == pid) {
```
jshkurti: === please (oui je cherche la petite bete mais c'est important ^_^)

soyuka: du coup je `parseInt(, 10)` ? :p

lib/God/Methods.js
```diff
@@ -221,11 +240,21 @@ module.exports = function(God) {
     if (!pid) return cb({msg : 'no pid passed or null'});
 
     try {
-      treekill(pid);
+      var p = God.findOneByPID(pid);
+
+      if(p === null || p.treekill === true) {
```
jshkurti: Je comprends pas la logique, si p === null ca veut dire que le pid n'existe pas dans la liste des process pm2, pourquoi faire un treekill alors ? :o

lib/God/Methods.js
jshkurti: ah, le pid peut etre une string ?

lib/God/Methods.js
```diff
@@ -221,11 +240,21 @@ module.exports = function(God) {
...
+        treekill(pid);
+      } else {
+        //if we don't want to kill the tree we just SIGINT the process
+        //sending SIGINT with the process group will kill all child processes but the detached ones
+        process.kill(-pid, 'SIGINT'); //- act as as process group
```
gcochard: Maybe refer to the [kill(2)](http://linux.die.net/man/2/kill) docs and/or say "negative means group..." rather than "-" here? At first glance this looks like a bug.

soyuka: Yes negative pid. See [this interesting answer](http://unix.stackexchange.com/questions/45426/why-would-ctrl-c-behave-differently-than-kill-2).

lib/God/Methods.js
soyuka: En fait le test est fais au dessus, le test sert un peut à rien c'est sûr :p

lib/God/Methods.js
soyuka: Sait-on jamais ? :) Je l'ai vu dans treekill le parseInt pour ça ^^.

lib/God/Methods.js
soyuka: ? If pid is less than -1, then sig is sent to every process in the process group whose ID is -pid.

ryanbmilbourne:
@soyuka Any update on this pull request? I see the build is still failing. :crying_cat_face:

soyuka:
I'm closing this and'll do a clean PR for the `fork_mode`.

see https://github.com/Unitech/PM2/issues/1036#issuecomment-91133953

PR #1120, opened 2015-03-24
Fix startup scripts for centos

kcormier:
Fixes the startup scripts for centos 7/systemd when running as an alternative user

Specifically addresses the issue that pm2 save/dump was being run as root rather than the correct user.

Also lays the groundwork for using the /etc/os-release file to determine which os/variant you are running on.  This seems to be something that many distros are moving toward to make this task easier.

I would be interested to know if there is a better way to get the user's home directory.  That felt particularly hacky to me.


package.json
```diff
@@ -167,6 +167,8 @@
     "shelljs"             : "0.3.0",
     "isbinaryfile"        : "^2.0.3",
     "semver"              : "4.3.1"
```
jorge-d: You're missing a coma in here :)

jorge-d:
Can you drop the `npm.install.dev` file ? Doesn't seem to belong in here :)

kcormier:
@jorge-d Updated.  Thanks for the heads up!

claflico:
I'm trying to puppetize the install and configuration of pm2 as a service on CentOS 7 and the deployment fails when I use the pm2-init-centos.sh file as a template. If I use the pm2.service file as the template the deployment doesn't fail but the service shows to be dead every time I do a status. Would this pull request fix that?

kcormier:
@claflico If I understand your issue correctly, then it should.  I'd be very curious to get feedback either way if you do try it.

kcormier:
Also, can anyone let me know what, if anything, is blocking this from being merged?

Thanks!

claflico:
@kcormier With a bit of trial and error I was able to figure out the proper way to use the pm2.service file. The issue I was running into was that the pm2 service would not start if the dump file did not exist in the .pm2 directory in the home directory of the user that pm2 is running as. This means that the app had to be started via command line first before the service would function. 

If I started my app, did a manual dump, then stopped and deleted the app I could then get the pm2 service to start, stop & restart properly and even start the app on reboot. However I ran into an issue with one testing box where every other restart the app would not start and when I looked the dump file was empty. I couldn't figure out the cause since it worked on another node reliably and both nodes were identical except for IP & name since they were puppet managed. 

I ended up using puppet to create an /etc/pm2/processes.json file with the info about the app so that I could rename the app and redirect the logs to the /var/log/pm2/ directory.....I then changed the pm2.service file to start/stop the processes.json file rather than resurrect and dump. 

I've had this method working for a week now and it appears to be reliable with puppet.

kcormier:
From the code, it looks like the intended behavior was for the dump to be run as part of pm2 startup.  When changing the user, however, the dump was not being put in the correct home dir.  This PR should fix that issue as well.

Issue #1121, opened 2015-03-24
init.d script PATH ordering


jstans:
In the case of having an out-dated system installed binary of node(/iojs) when wanting to use an nvm installed binary, the auto-generated pm2 startup script will find the system installed binary first and try and use it.

Suggest simply moving the ordering so the script puts the binary PATH at the beginning.

jshkurti:
Merged.

PR #1122, opened 2015-03-24
Use NODE_PATH before system paths

jstans:
Fix System Path being used before NODE_PATH, to override system installed versions of node. #1121


jshkurti:
Hmm why did you closed it ?

jshkurti:
Alright

PR #1125, opened 2015-03-25
fix #1124 wrong handling of PM2_PROGRAMMATIC flag

### LINKED ISSUE #1124: Wrong operation of environment variable.
I think that this commit d5e731e3 is wrong operation of environment variable.

Example: https://github.com/Unitech/PM2/blob/0.12.8/lib/CLI.js#L961
`process.env.PM2_PROGRAMMATIC === true`
It absolutely not true, because environment variable is string.

ww24:


jshkurti:
@Unitech

Unitech:
small tweak, sounds good to me, thanks!

Unitech:
merged

PR #1141, opened 2015-03-31
use bash shell by default in pm2 ubuntu init script

adamhadani:
Hey,
In a recent attempt to use PM2 on an Ubuntu instance (on an AWS EC2 instance), using the startup script failed for me. After some probing around i identified the problem was that when using pm2 with a system user for startup, i.e invoking:
```
        $ pm2 startup ubuntu -u www-data
```
The resulting init script would fail to execute correctly due to www-data having no login by default (by default on Ubuntu its shell is set to /usr/sbin/nologin). 

changing the super() command to invoke su with -s /bin/bash seems to work and I cannot see any scenario where this would be a problem in this context.

Attached the PR for this issue, happy to iterate on this with you guys

thanks,
Adam

### DIFF
--- a/lib/scripts/pm2-init.sh
+++ b/lib/scripts/pm2-init.sh
@@ -19,13 +19,24 @@
 NAME=pm2
 PM2=%PM2_PATH%
 USER=%USER%
-SHELL=/bin/bash
 
 export PATH=%NODE_PATH%:$PATH
 export PM2_HOME="%HOME_PATH%"
 
+get_user_shell() {
+    local shell=$(getent passwd $1 | cut -d: -f7)
+
+    if [[ $shell == *"/sbin/nologin" ]] || [[ $shell == "/bin/false" ]];
+    then
+      shell="/bin/bash"
+    fi
+
+    echo "$shell"
+}
+
 super() {
-    su - $USER -s $SHELL -c "PATH=$PATH; $*"
+    local shell=$(get_user_shell $USER)
+    su - $USER -s $shell -c "PATH=$PATH; $*"
 }
 
 start() {
--- a/lib/scripts/pm2-init.sh
+++ b/lib/scripts/pm2-init.sh
@@ -19,12 +19,13 @@
 NAME=pm2
 PM2=%PM2_PATH%
 USER=%USER%
+SHELL=/bin/bash
 
 export PATH=%NODE_PATH%:$PATH
 export PM2_HOME="%HOME_PATH%"
 
 super() {
-    su - $USER -c "PATH=$PATH; $*"
+    su - $USER -s $SHELL -c "PATH=$PATH; $*"
 }
 
 start() {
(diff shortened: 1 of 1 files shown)


soyuka:
Lgtm !
@jshkurti I let you do the merge it's on master

neonstalwart:
does this affect cases where someone may, for example, be using a user with a default zsh shell that sets up some parts of their environment using .zshenv, etc?

adamhadani:
@neonstalwart @soyuka -  seems like a pretty rare scenario to me for someone to opt using zsh on Ubuntu, and then create a 'safe user' for running pm2 and specify zsh for that user, and then add stuff to that safe user's env file.. but if you guys wanna be ultra-conservative, we can do something like this which for the price of added verbiage would work for ubuntu / *nix where getent is available:

```
get_user_shell() {
    local shell=$(getent passwd $1 | cut -d: -f7)

    if [[ $shell == *"/sbin/nologin" ]] || [[ $shell == "/bin/false" ]];
    then
      shell="/bin/bash"
    fi

    echo "$shell"
}

super() {
    local shell=$(get_user_shell $USER)
    su - $USER -s $shell -c "PATH=$PATH; $*"
}
```

neonstalwart:
> seems like a pretty rare scenario to me...

@adamhadani it's not rare for me :smile: i use zshenv to source nvm.sh so that even non-interactive shells will have node available.

adamhadani:
@neonstalwart  - got it, I take it back, apologies sir! ;)
let me know if you guys think that last piece of code seems legit then, and i'll create new PR

soyuka:
I'm using `/etc/profile.d` and sourcing it on the `bashrc` or `zshrc` if
it's a global env variable. If it's specific I usually add it to my
`pm2.json` or calling it on execution `ENV_VAR=something pm2 start xx.js`.

@adamhadani the work around seems fine to me too!
Le mer. 1 avr. 2015 à 20:10, Adam Ever-Hadani <notifications@github.com> a
écrit :

adamhadani:
@soyuka  - got it, cool. I just committed that workaround to the same branch, looks like this PR is getting automagically updated with that commit (85fb58c)

soyuka:
Ok just tried it and it's working fine. Some thoughts, if the user is not defined, should we add the use of `whoami`?

adamhadani:
@soyuka , looking at lib/CLI.js, it looks like %USER% will be always interpolated into a non-null value (either user-supplied paramter or 'root'), but I added this in anyhow as a default inside get_default_user() in case no parameter is passed, just committed

soyuka:
@jshkurti I think we can merge this! 

@adamhadani should we do the same for every other platform that has an upstart script?

adamhadani:
@soyuka - I would say, this can be done and probably should be done for any platform that 
1. supports getent command or similar method to get a user's login shell
2. advocates usage of no login shell for service / system users 

I see there's a centos and amazon flavors of the script - I don't know enough personally about either one to make such assumptions but if you guys know better, it should be straightforward exercise to port this modification to those as well

PR #1156, opened 2015-04-08
Updated init script for Amazon and CentOS

### LINKED ISSUE #1100: Amazon EC2 : saved PM2 configuration is not recognized at startup (when running pm2 in user mode)
Scenario 

```
// create an EC2 instance, install node, install pm2
// create an app (install sails for example)
// add your app to pm2
[ec2-user]$ pm2 start app.js -n <myapp> -x -- --prod
// configure PM2 startup as running in user mode (-u <user>)
[ec2-user]$ sudo env PATH=$PATH:/usr/bin pm2 startup amazon -u ec2-user
// save configuration
[ec2-user]$ pm2 save
=> OK :  writes current PM2 configuration (my app) into  ec2-user home dir, 
=> OK the file ~/.pm2/dump.pm2 is generated
// go to your AWS console, reboot your EC2 instance
=> FAILS : after restart, <myapp> is not started
```

EgoAleSum:
This fixes #1100

The script contains the fixes from the "centos" one to change user.
In addition, contains @ObjectIsAdvantag 's fix for PM2_HOME

### DIFF
--- a/lib/scripts/pm2-init-amazon.sh
+++ b/lib/scripts/pm2-init-amazon.sh
@@ -24,7 +24,6 @@ PM2=%PM2_PATH%
 USER=%USER%
 
 export PATH=%NODE_PATH%:$PATH
-export PATH=$PATH:%NODE_PATH%
 if [[ "$USER" == "root" ]]; then
     export PM2_HOME="%HOME_PATH%"
 else
--- a/lib/scripts/pm2-init-centos.sh
+++ b/lib/scripts/pm2-init-centos.sh
@@ -24,7 +24,11 @@ PM2=%PM2_PATH%
 USER=%USER%
 
 export PATH=%NODE_PATH%:$PATH
-export PM2_HOME="%HOME_PATH%"
+if [[ "$USER" == "root" ]]; then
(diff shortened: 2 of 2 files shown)


soyuka:
Could you please take a look at #1120 (centos startup fix) to see if it corrects the same issue you had regarding centos? If not, please comment on there so we get along with those pull requests, thanks!

claflico:
What version of CentOS would this be for? I'm trying to puppetize the install and configuration of pm2 as a service on CentOS 7 and the deployment fails when I use the pm2-init-centos.sh file as a template.

EgoAleSum:
@claflico what error do you get?

claflico:
Since CentOS 7 uses systemd Puppet wants to use "systemctl enable name" to define a service and that fails because it can't find the /usr/lib/systemd/system/pm2.service file. I was able to resolve it by doing away with using the chckonfig/pm2-init-centos.sh and using the pm2.service file instead.

Unitech:
https://github.com/Unitech/pm2/blob/master/lib/templates/init-scripts/upstart.tpl#L18
Upgrade to latest PM2

PR #1173, opened 2015-04-14
new command: pm2 gc (triggers garbage collection)

jshkurti:
You can now ```pm2 gc``` to manually trigger garbage collection for the PM2 daemon process.

### DIFF
--- a/ADVANCED_README.md
+++ b/ADVANCED_README.md
@@ -320,7 +320,7 @@ JSON:
 Units can be K(ilobyte), M(egabyte), G(igabyte).
 
 >
-Actually the way it works when you type : 
+Actually the way it works when you type :
 `pm2 start app.js --max-memory-restart 50M`
 commander module will transform it to `maxMemoryRestart: "50M"`, then after being processed by PM2 logic it will become an env variable as follows `max_memory_restart : 52428800 // in bytes this time`.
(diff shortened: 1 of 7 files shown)


soyuka:
Is `global.gc()` available to every node version? We could add a log entry
if it's not available?

I think that you could update the docs too :). Sry, wanted to comment on
the code but I'm on the phone right now :).
Le mar. 14 avr. 2015 à 19:42, Joni Shkurti <notifications@github.com> a
écrit :

jshkurti:
I read about it in a blog post from 2013 and it still works on node 0.12.2 so I guess it's been there for a long time :)

soyuka:
Just noticed the `if` :)
Le mar. 14 avr. 2015 à 20:10, Joni Shkurti <notifications@github.com> a
écrit :

jshkurti:
After reading this http://www.jayway.com/2015/04/13/600k-concurrent-websocket-connections-on-aws-using-node-js/

- Added some more flags to node.js.
- Added some documentation
- Added some tests
- Now gc runs in the pm2 worker every 30seconds

jshkurti:
it only concerns PM2 daemon itself for now.
Not the users' apps :)

soyuka:
Oh my bad!
Le mar. 14 avr. 2015 à 21:09, Joni Shkurti <notifications@github.com> a
écrit :

PR #1176, opened 2015-04-15
Man entry

soyuka:
It's not perfect but it's a start :).


jshkurti:
Nice :+1:

soyuka:
Generated with: https://github.com/sunaku/md2man

We could add more man pages through the package.json.

Issue #1109, opened 2015-03-19
PM2 logs should show errors in the order in which they appear


murphyj:
Noticed that running pm2 logs pushes restart sequences to the top of the logs for a pm2 process but errors to the bottom. e.g.

```
########### Starting streaming logs for [all] process
PM2: 2015-03-19 10:46:34: App closed with code: null
PM2: 2015-03-19 10:46:34: App name:app id:0 exited
PM2: 2015-03-19 10:46:34: Process with pid 25017 killed
PM2: 2015-03-19 10:46:34: Starting execution sequence in -fork mode- for app name:app id:0
PM2: 2015-03-19 10:46:34: App name:app id:0 online
PM2: 2015-03-19 10:46:54: App closed with code: null
PM2: 2015-03-19 10:46:54: App name:app id:0 exited
PM2: 2015-03-19 10:46:54: Process with pid 25186 killed
PM2: 2015-03-19 10:46:54: Starting execution sequence in -fork mode- for app name:app id:0
PM2: 2015-03-19 10:46:54: App name:app id:0 online
PM2: 2015-03-19 10:49:09: App closed with code: null
PM2: 2015-03-19 10:49:09: App name:app id:0 exited
PM2: 2015-03-19 10:49:09: Process with pid 25260 killed
PM2: 2015-03-19 10:49:09: Starting execution sequence in -fork mode- for app name:app id:0
PM2: 2015-03-19 10:49:09: App name:app id:0 online
PM2: 2015-03-19 10:53:12: App closed with code: null
PM2: 2015-03-19 10:53:12: App name:app id:0 exited
PM2: 2015-03-19 10:53:12: Process with pid 27263 killed
PM2: 2015-03-19 10:53:12: Starting execution sequence in -fork mode- for app name:app id:0
PM2: 2015-03-19 10:53:12: App name:app id:0 online
app-0 (err):     at Module._compile (module.js:460:26)
var express = require('express'),
app-0 (err):     at Object.Module._extensions..js (module.js:478:10)
app-0 (err):     at Module.load (module.js:355:32)
app-0 (err):     at Function.Module._load (module.js:310:12)
app-0 (err):     at Function.Module.runMain (module.js:501:10)
app-0 (err):     at startup (node.js:129:16)
app-0 (err):     at node.js:814:3
app-0 (err): /var/lib/jenkins/morph-tools/webhooks/node_modules/express/lib/express.js:89
[...]

PR #1187, opened 2015-04-17
Sort logs by modified time asc.

Tjatse:
#1109


Tjatse:
Any ideas?

jshkurti:
I don't think this is really important. In any case, showing the last 20 lines of error logs and then the last 20 lines of output logs (or the other way around) is still confusing for the user.
I've been myself confused many times.
That, or the fact that sometimes it shows logs from completely different apps just because they had the same name as my actual app.

The solution to the logs order issue would be to always merge logs and then show the merged logs file.

I think logs should be merged by default and seperated in two different files only if the user specifies it.

Tjatse:
It's not that complicated.
For example, the `stderr` logs were recorded 2 minutes ago, and the `stdout` logs are being recorded every seconds, but when you running `pm2 logs`, the outputs are:
```
[xxx.xx] stdout 1....
[xxx.xx] stdout 2....
[xxx.xx] stdout 3....
[xxx.xx] Error stack:
    from xxx.js line xx....
    function xxx() throw error...
    ....
```

The `stderr` always at the end of logs, that's unreasonable.

jshkurti:
Hey @Tjatse 
I'm completely refactoring `pm2 logs`.
The purpose of this refactor is to get rid of the spawn('tail') and only use natives modules like `fs` to tail logs from files and then use the `bus` event emitter to grab real-time logs.

This way, when you launch a new app while `pm2 logs` is running, you don't have to Ctrl^C and `pm2 logs` again. It just catches the new logs.

And I'm also implementing the sorted by modified time feature.

Here, take a look if you got time : https://github.com/Unitech/PM2/pull/1341

Tjatse:
gocha! it seems great, nice job.

Issue #1195, opened 2015-04-19
Startup script fails to run because get_user_shell returns nothing


buunguyen:
PM2 couldn't resurrect processes after reboot in a stock Rackspace Ubuntu box. From debugging pm2-init.sh, `get_user_shell` always prints empty, causing the command `su - $USER -s $shell -c "PATH=$PATH; $*"` to fail. I temporarily hacked around by using `/bin/sh` directly, i.e. replacing the above "$shell" with "/bin/sh".  Then reboot & resurrection work.

Is this a bug of the init script or should I configure something in the server beforehand? 

**Environment:**
Ubuntu 14.04.2
PM2 0.12.11
iojs 1.7.1

jshkurti:
Hey @adamhadani, can you have a look at this please ? :)

adamhadani:
Sure,
@buunguyen - could you please try to invoke the following two commands on your terminal using the same user you try to use the pm2 script with and see what you get back ?

$ whoami
[should echo current logged in user]
$ getent passwd $USER
[should echo a long line from /etc/passwd for current user]

I'm thinking anyway probably safest way to go, unless we find some specific issue here, is adding an if statement to get_user_shell that defaults to something like '/bin/bash' in case $shell is empty before echoing it for whatever reason..

buunguyen:
@adamhadani both commands return the "right" data running as the user starting pm2. However, for some reason, `/etc/passwd` doesn't include the shell in the line corresponding to this user. Probably because I forgot to specify the shell when creating the user.

I don't know if there's any problem defaulting to /bin/bash, but I think it's probably a good idea. But at least, it should be useful printing out some warning/error so that users can check their /etc/passwd instead of just executing the statement with an empty $shell.

Thanks for checking out.

jshkurti:
@adamhadani, if you could do a PR for this it would be great ! :+1:

buunguyen:
Certainly, will submit one.

adamhadani:
@jshkurti @buunguyen https://github.com/Unitech/PM2/pull/1199

animedbz16:
I believe that this issue would also fix #1061, as centrify authentication is done from a central authority and user accounts and passwords are not directly managed in /etc/passwd on the machine you are on.

PR #1199, opened 2015-04-20
support use case of user's shell being empty in /etc/passwd for pm2-init...

adamhadani:
See Issue #1195


jshkurti:
Great! Thanks! :D

buunguyen:
:+1: glad it's in your hand, much better than anything I came up with.

PR #1201, opened 2015-04-21
Added flag: pm2 logs --raw

jshkurti:
`--raw` flag shows logs without any timestamp or other PM2 related metadata.


lib/CLI.js
```diff
@@ -1363,9 +1363,11 @@ function fallbackLogStream(id, lines) {
       exitCli(cst.ERROR_EXIT);
     }
 
-    Log.stream(cst.PM2_LOG_FILE_PATH, lines);
+    if (!raw)
+      Log.stream(cst.PM2_LOG_FILE_PATH, lines);
 
-    printOut('########### Starting streaming logs for [%s] process', id || 'all');
+    if (!raw)
```
jorge-d: Why not merge this `if` statement with the one two lines above?

jshkurti: Indeed :+1:

Issue #1193, opened 2015-04-18
startup script per user


ashconnell:
I have a project that run staging and production on a single digital ocean droplet. Staging and production both have their own users. The `pm2 startup` command as one of the users creates a init.d script, and running it as another user just overwrites the first script.

PierreGambarotto:
The PR #1202 can solve this problem by adding the username to the init script name. Hope this helps.

Unitech:
https://github.com/Unitech/pm2/issues/2559

Issue #889, opened 2014-12-19
`pm2 startup` cannot generate startup scripts for multiple users


dantman:
When using `pm2 startup` (at least when using `pm2 startup ubuntu -u x`) the same init.d file is used every time.

This means that because `--run-as-user` has been removed you can no longer easily have scripts started by pm2 that run under different users, they all have to run as the same user.

Deminetix:
+1

PierreGambarotto:
With the PR #1202, `pm2 startup -u username …` adds the username to the init script name.
Hope this helps.

DeveloperAlex:
I had the same problem on ubuntu 14.04.  Fixed it by running "sudo pm2 startup -u myusername".   Then "sudo pm2 save".  And "sudo shutdown -r now" - of course.

Unitech:
Closing in favor of #2559 

To try the new startup system:

```bash
$ npm install Unitech/pm2#development -g
$ pm2 update
```

```bash
$ pm2 startup     # Then copy / paste the command displayed
$ pm2 unstartup
```

PR #1202, opened 2015-04-23
adds opts.user to the init script name

PierreGambarotto:
This very simple PR adds the user name, from the `-u --user` switch to the name of the init script generated by `pm2 startup …`.

This should contribute a possible solution to issues  #889 #1193

### DIFF
--- a/lib/CLI.js
+++ b/lib/CLI.js
@@ -439,6 +439,10 @@ CLI.startup = function(platform, opts, cb) {
 
   var user = opts.user || 'root';
 
+  if (user != 'root'){
+    scriptFile = scriptFile.replace(/pm2/i, '$&-'+user);
+  }
+  
   script = script.replace(/%PM2_PATH%/g, process.mainModule.filename)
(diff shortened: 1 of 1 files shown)


Deminetix:
Looks great, is it a full proof solution or only partially?

PierreGambarotto:
@Deminetix 
It is a very simple workaround. To give an exemple, I'll use the case of an ubuntu, but the reasoning is the same regardless the OS.

By default, `pm2 startup` generates `/etc/init.d/pm2-init.sh`, in the case of an ubuntu.
The proposed modification simply adds the userame, thus `pm2 startup -u foo` now generates `/etc/init.d/pm2-foo-init.sh`.

If no `-u/--user` switch is used, then the previous usage remains, and `/etc/init.d/pm2-init.sh` is generated.

jshkurti:
Looks ok to me.
@Unitech what do you think ?

creynders:
:+1:

vmarchaud:
Done in PM2 `2.4`

PR #1213, opened 2015-04-27
Set PM2_HOME on the user session starting pm2.

miguelcnf:
This is a quick fix to update the majority of the init scripts to set the PM2_HOME environmental variable to whatever it's defined on the session of the user running pm2.

Not having this caused $PM2_HOME to default to $HOME/.pm2 which might not be consistent and break the startup ability - since dump was created in a different location.

It also ports the run as user ability from the centos init script to the amazon init script.

### DIFF
--- a/lib/scripts/pm2
+++ b/lib/scripts/pm2
@@ -9,7 +9,7 @@ export PATH=%NODE_PATH%:$PATH
 export PM2_HOME="%HOME_PATH%"
 
 super() {
-    su - $USER -c "PATH=$PATH; $*"
+    su - $USER -c "PATH=$PATH; PM2_HOME=$PM2_HOME $*"
 }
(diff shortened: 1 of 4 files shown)


soyuka:
lgtm! Basically adds `$PM2_HOME` to environment variables.

jarosser06:
:+1:

miguelcnf:
Without the change on the amazon init script persisting the running as a different user config over reboots is not supported on amazon linux even if PM2_HOME is set to the default location $HOME/.pm2.

miguelcnf:
Any feedback on this?

jshkurti:
Thanks @miguelcnf :)

miguelcnf:
Awesome - thanks!

PR #1220, opened 2015-04-29
Wrapped fork_mode

jshkurti:


Unitech:
Check that `node app.js` and `pm2 start app.js` has the same values for the variables:

- NODE_PATH
- __filename
- __dirname
- module
- PWD
- require.main.filename

Once we have a test covering this we are good to merge the wrapped fork

Unitech:
There is a kind of problem, as we inject pmx, and custom probes send data periodically to PM2, the process does not exit automatically. The behavior is the same in cluster mode. It's a kind of breaking change.

https://gist.github.com/Unitech/e46e0d914d0f9c62f739

To keep the same behavior than before I suggest to deactive [custome probes](https://github.com/keymetrics/pmx/blob/master/lib/index.js#L49) in fork mode

thoughts?

jshkurti:
I tried desactivating custom probes, the process still remains alive.
The only solution for it not to be is to not inject pmx at all.

Actually, I think keeping the process alive is more natural.
A lot of users just want to test pm2 with a simple script `console.log('hello world')` and they don't understand why it prints 15 times 'hello world' to the output.
If an exception occurs or an process.exit(), the behaviour is still the same.
I don't think this is a problem.

PR #1222, opened 2015-04-30
Use USER constant in pm2-init-amazon.sh

mori-dev:
These are similar issues.

- https://github.com/Unitech/PM2/issues/1100
- https://github.com/Unitech/PM2/pull/1156

In [this](https://github.com/Unitech/PM2/commit/602677f5e8ab3e29c3e3eb8ed729c3a37763d437) commit, 'super()' is omitted Intentionally. As a result, -u option of pm2 command cannot be recognized. I think we should use 'super()'.


Unitech:
+1

soyuka:
https://github.com/Unitech/PM2/commit/4ffa70c720cc32af53847ef35914601b45196609

PR #1237, opened 2015-05-05
Use svg-badges in README

SimenB:
Much higher quality.

image

vs.

image

(Even higher quality when not pasting screenshot into GitHub...)
https://github.com/SimenB/PM2/blob/badges/README.md


jshkurti:
Thanks

Issue #1260, opened 2015-05-12
env args superseed ecosystem.json args


thunderace:
When args is set in starting shell env, ecosystem args is ignored.
Example : 
shell env : 
thunder@debianServer:/media/data/home/thunder$ env
MAIL=/var/mail/thunder
../..
vizion=true
HOME=/home/thunder
SHLVL=5
args=-b,-p,3132,-w,/media/data/home/thunder,-l,0.0.0.0,-a,:
../..

with this domo.json file : 
{
  "apps" : [{
    "name"        : "ArDomo2",
    "script"      : "nodes/domoSerial/domoSerial.js",
    "args"        : "['--node=ArDomo2']",
    "cwd"         : "/home/thunder/repo.git/domo.mqtt",
    "log_date_format"  : "YYYY-MM-DD HH:mm Z",
    "pid_file"   : "/home/thunder/.pm2/pids/ArDomo2.pid",
    "error_file" : "/home/thunder/.pm2/logs/ArDomo2-err.log",
    "out_file"   : "/home/thunder/.pm2/logs/ArDomo2-out.log",
    "merge_logs"  : true
  }]
}
pm2 start domo.json 
domoSerial.js recieve these args : -b,-p,3132,-w,/media/data/home/thunder,-l,0.0.0.0,-a,: and not --node=ArDomo2

jshkurti:
Can you try it with `npm i -g git://github.com/Unitech/PM2#development` ?
I think this might have been fixed.

thunderace:
OK, i will try this this week-end.
Thank you

thunderace:
Fixed, thank you

PR #1282, opened 2015-05-18
closes #941: PM2 executed from within node inherits command line arguments

### LINKED ISSUE #941: PM2 executed from within node inherits command line arguments
This is a weird one. Hopefully I can explain clearly. Let's say that I run `pm2 start a.js -- 3000` and _a.js_ calls `require('child_process').exec('pm2 start b.js -- 4000')` then _b.js_ will receive 3000 as it's argument instead of 4000, as expected. This is true even if _a.js_ uses the API, e.g. `pm2.start('b.js', {rawArgs: ['--', 4000]})`. The child process seems to inherit the command line arguments of the parent.

boneskull:
Hi,

I'm unfamiliar with this codebase, so I'm thinking the test I wrote for this was pretty roundabout.  Would appreciate any suggestions.

The `should list processes` test is failing for me locally with or without my changes.

### DIFF
--- a/lib/Common.js
+++ b/lib/Common.js
@@ -321,7 +321,7 @@ Common.extend = function(origin, add){
   if (!add || typeof add != 'object') return origin;
 
   //Ignore PM2's set environment variables from the nested env
-  var keysToIgnore = ['name', 'exec_mode', 'env', 'pm_cwd', 'exec_interpreter', 'pm_exec_path', 'node_args', 'pm_out_log_path', 'pm_err_log_path', 'pm_pid_path', 'pm_id', 'status', 'pm_uptime', 'created_at', 'started_inside', 'unstable_restarts', 'restart_time', 'pm_id', 'axm_actions', 'pmx_module', 'command', 'watch', 'versioning', 'vizion_runing', 'MODULE_DEBUG'];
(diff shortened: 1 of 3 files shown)


jshkurti:
Hey, thanks a lot man. :+1: 
I'm just going to grab the changes on lib/Common.js for now and publish PM2 v0.12.15.
We'll see about the tests failing issue later.

jshkurti:
Do you think it also fix this one ? https://github.com/Unitech/PM2/issues/1260
And this ? https://github.com/Unitech/PM2/issues/1283
It would be great

boneskull:
I'll take a look later. Out of curiosity, how would you write a test for this?  the programmatic API didn't have a whole lot of documentation, so actually asserting failure within a test took awhile for me to figure out--I'm sure there's an easier way than using `bus`.

boneskull:
Also it wasn't clear how to debug a running daemon.

boneskull:
Regarding the other two issues, I'm not sure without reproducing them in a test case.  

But from looking around in there, the *real* problem is [right here](https://github.com/Unitech/PM2/blob/master/lib/God.js#L63), where `God.executeApp()` does this [`Common.extend()`](https://github.com/Unitech/PM2/blob/master/lib/Common.js#L319) call.  Because this is a "shallow" extend, any child objects of `env_copy` will simply be overwritten.  This was known, because of the [`keysToIgnore`](https://github.com/Unitech/PM2/blob/master/lib/Common.js#L324) variable.  But I'm unsure if any of this is necessary.

What you may want instead is a "deep extend" or "merge" -- for an example, see [_.merge](https://lodash.com/docs#merge).  If you want to include it without including the entirety of `lodash` (wouldn't blame you), you can `npm install --save lodash.merge`, and `var merge = require('lodash.merge')`.

Doing this would likely be a breaking change, as child processes will have *more stuff* in their environment.  It seems "more correct" to me, however.

PR #1298, opened 2015-05-23
Support JS in addition to JSON and JSON5 config files

### LINKED ISSUE #1296: Request: allow ecosystem.js config files in addition to json and json5
I would like to be able to include a little bit of code in my config files for flexibility and readability. For example, the deploy section of my config file currently looks like this:
```js
  deploy: {
    production: {
      ...
      'post-deploy': 'npm install && npm install git+ssh@github.com:keys/example.git && gulp babel && pm2 startOrRestart ecosystem.json5 --env production',
    }
  }
```

I would like to be able to write:
```js
  deploy: {
    production: {
      ...
      'post-deploy': [
        'npm install',
        'npm install git+ssh@github.com:keys/example.git',
        'gulp babel',
        'pm2 startOrRestart '+ path.basename(__filename) + ' --env production',
      ].join(' && ')
    }
  }
```

ide:
This adds support for full JS within ecosystem config files using the sandboxing capability of Node's vm module. Since JSON and JSON5 are both subsets of JS and don't access any variables in the JS environment, there is no change to existing config files.

Tested with Node 0.10, Node 0.12, io.js 1.8, and io.js 2.0.

Fix #1296

### DIFF
--- a/lib/CLI.js
+++ b/lib/CLI.js
@@ -6,7 +6,9 @@ var fs                   = require('fs');
 var path                 = require('path');
 var async                = require('async');
 var debug                = require('debug')('pm2:monit');
+var semver               = require('semver');
 var util                 = require('util');
+var vm                   = require('vm');
 var chalk                = require('chalk');
 var exec                 = require('child_process').exec;
 
@@ -214,7 +216,7 @@ CLI.deploy = function(file, commands, cb) {
     env = args[1];
 
   try {
-    var json_conf = json5.parse(fs.readFileSync(file));
+    var json_conf = parseConfig(fs.readFileSync(file), file);
   } catch (e) {
     printError(e);
     return cb ? cb(e) : exitCli(cst.ERROR_EXIT);
@@ -844,7 +846,7 @@ CLI.remote = function(command, opts, cb) {
 /**
  * Start or restart|reload|gracefulReload a JSON configuration file
  * @param {string} action    restart|reload
- * @param {string} json_conf json file path
+ * @param {string} json_conf JS file path
  * @param {string} opts      option like environment type and co
  * @callback cb optional
  * @param cb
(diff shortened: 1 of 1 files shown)


Unitech:
Awesome! Will review/merge it by tomorrow

ide:
Hi @Unitech! Have you had a chance to review my commit?

Unitech:
I did a weird merge, so this PR has not been notified. I made some tests and fixed some other stuff, it's already available in the latest version of PM2 (0.12.16) ;) [+added as contributor](https://github.com/Unitech/PM2/blob/development/package.json#L58) 

If you have any other features in mind feel free to PR!

Best,
Alex

ide:
Thank you for the updating and for adding tests =)

aaronroberson:
I'm using pm2 0.10.10 and this doesn't appear to be working. If I use a js file with a config in it pm2 treats the js file as if it were my application. Here is my `startup.js` file:

```
{
  apps: [{
    name: "app",
    script: "/var/www/app/current/server.js",
    exec_mode: "cluster_mode",
    instances: -1,
    max_memory_restart: "1G",
    max_restarts: 10,
    watch: true,
    watch_options: {
      followSymLinks: true
    }
  }]
}
```

When I run `pm2 list` the app is running as `startup.js` instead of `app` and the script path is to the `statup.js` file instead of the `server.js` file in the config.

Unitech:
Upgrade to the latest PM2

aaronroberson:
0.15.10 is the latest available version via npm.

ide:
It might be the file extension. Try naming the file ecosystem.json or json5.

aaronroberson:
I don't want to use a JSON file I want to use a JavaScript file with a config in it as this issue and the documentations suggests I can do. @ide are you suggesting I name my JavaScript file with a .json extension even though it is not valid JSON?

ide:
Yes I believe it's maybe just a cosmetic issue with the file extension.

PR #1341, opened 2015-06-10
`pm2 logs` completely refactored

jshkurti:


lib/Log.js
```diff
@@ -1,104 +1,77 @@
...
-    this.count ++;
-    if(this.count >= this.colors.length){
-      this.count = 0;
+Log.tail = function(apps_list, lines, raw, callback) {
+  if (lines === 0 || apps_list.length === 0)
+    return callback && callback();
+
+  var count = 0;
+
+  apps_list.forEach(function(app, index) {
```
Tjatse: The apps_list still should be sorted by `mtime` of `app.path`, otherwise the #1109 issue still exists, what do you think?

SOL:
```javascript
!!apps_list.length && apps_list.sort(function(a, b){
  return (fs.existsSync(a.path) ? fs.statSync(a.path).mtime.valueOf() : 0) -
(fs.existsSync(b.path) ? fs.statSync(b.path).mtime.valueOf() : 0);
}).forEach('[YOUR CODE GOES HERE]');
```

Tjatse:
If the `bus:log` does not cause any performance (Socket traffic, memory, blabla....) problem, this PR seems very nice!

lib/Log.js
jshkurti: That's exactly what I was going to do ^^
You just made the job easier :D
Thanks

lib/CLI.js
```diff
@@ -1554,85 +1554,91 @@ CLI.monit = function(cb) {
...
-      // Ignored meaningless logs.
-      if (!pk || pk.toLowerCase() == '/dev/null') {
-        return;
+    // Populate the array `files_list` with the paths of all files we need to tail
+    list.forEach(function(proc) {
+      if (proc.pm2_env
+         && (id === 'all'
+            || proc.pm2_env.name == id
+            || proc.pm2_env.pm_id == id)) {
+        if (proc.pm2_env.pm_out_log_path)
```
Tjatse: AFAIC, I don't like dry codes :)
It's difficult to maintain copies.
How about do it in this way:
```javascript
['out', 'err'].forEach(function(lt){
  var _path;
  if(_path = proc.pm2_env['pm_' + lt + '_log_path']){
    pushIfUnique({
      path     : _path,
      app_name : proc.pm2_env.name + '-' + proc.pm2_env.pm_id,
      type     : lt
    });
  }
});
```

lib/CLI.js
```diff
@@ -1554,85 +1554,91 @@ CLI.monit = function(cb) {
...
-      if (first.pm2_env.merge_logs) {
-        // Join `pm_id`s together.
-        first.pm_id = ps.map(function(proc){
-          return proc.pm_id;
-        }).join(',');
-        printLogs(id, lines, first, raw);
-      } else {
-        ps.forEach(function(ps0) {
-          printLogs(id, lines, ps0, raw);
+    if (!raw && (id === 'all' || id === 'PM2')) {
```
Tjatse: NO DRY CODES?
```
var tailAppLogs = function(){
  Log.tail(files_list, lines, raw, function() {
    if (!raw)
    printOut('[PM2] Streaming realtime logs for [%s] process%s', id, id === 'all' ? 'es' : '', '\n');
    Log.stream(id, raw);
  });
};
if (!raw && (id === 'all' || id === 'PM2')) {
  Log.tail([{
    path     : cst.PM2_LOG_FILE_PATH,
    app_name : 'PM2',
    type     : 'PM2'
  }], lines, raw, tailAppLogs);
  });
} else {
  tailAppLogs();
}
```

lib/Log.js
Tjatse: LOL.

PR #1384, opened 2015-06-24
Autocompletion based on tabtab 

alavit-d:
commander.commands is not mapped before the **completion** check, therefore no base command is avalaible

### DIFF
--- a/ADVANCED_README.md
+++ b/ADVANCED_README.md
@@ -12,6 +12,7 @@
 - [Options](#a987)
   - [Schema](#schema)
 - [How to update PM2?](#update-pm2)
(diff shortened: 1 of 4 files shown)


lib/scripts/sysCheck.js
```diff
@@ -0,0 +1,6 @@
+var exec = require('child_process').exec;
+
+if (/win32/.test(process.platform))
+  process.exit(0);
+exec('sh ./lib/scripts/completion.sh', function(err, stdout, stderr){});
+exec('node ./lib/scripts/keymetrics.js', function(err, stdout, stderr){ console.log(stdout); });
```
jshkurti: here ?

PR #1395, opened 2015-06-26
Add a --no-treekill option to avoid killing detached processes

soyuka:
I really need this because I'm handling detached processes in my main process.


jshkurti:
:+1:

PR #1403, opened 2015-06-29
Truncated output fix

jshkurti:
- Fixes the issue on CentOS when the output of `pm2 list` was truncated.


jshkurti:
@Tjatse Looks fine to me. I created a PR to see how it handles tests.
Your code is always complicated ^^

jshkurti:
I tested with iojs and Node.js, it works fine :)

Tjatse:
-;)

PR #1423, opened 2015-07-08
Update README.md

joshhornby:
Warning from the console: [WARN] Deprecated, we recommend using 0 instead of max to indicate maximum of instances

This PR fixes this deprecation.


jorge-d:
Thanks !

PR #1446, opened 2015-07-19
Create pm2.freebsd

jasin:
RC.script file for FreeBSD. This needs to go in /etc/rc.d/ and then you need to edit /etc/rc.conf and add these two lines

pm2_enable="yes"
pm2_user="" <-- user that you want pm2 to be ran as, default root user

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -45,6 +45,7 @@ var csts = {
   AMAZON_STARTUP_SCRIPT  : '../lib/scripts/pm2-init-amazon.sh',
   GENTOO_STARTUP_SCRIPT  : '../lib/scripts/pm2',
   DARWIN_STARTUP_SCRIPT  : '../lib/scripts/io.keymetrics.PM2.plist',
+  FREEBSD_STARTUP_SCRIPT : '../lib/scripts/pm2-freebsd.sh',
 
   SUCCESS_EXIT           : 0,
   ERROR_EXIT             : 1,
--- a/lib/CLI.js
+++ b/lib/CLI.js
@@ -573,9 +573,11 @@ CLI.startup = function(platform, opts, cb) {
     scriptFile = '/etc/systemd/system/pm2.service';
   }else if (platform == 'darwin') {
     scriptFile = path.join(process.env.HOME, 'Library/LaunchAgents/io.keymetrics.PM2.plist');
+  }else if (platform == 'freebsd') {
+    scriptFile = '/etc/rc.d/pm2';
   }
 
-  if(!!~['systemd', 'centos', 'amazon', 'gentoo', 'darwin'].indexOf(platform)){
+  if(!!~['freebsd', 'systemd', 'centos', 'amazon', 'gentoo', 'darwin'].indexOf(platform)){
     script = cst[platform.toUpperCase() + '_STARTUP_SCRIPT'];
   }
 
@@ -625,14 +627,20 @@ CLI.startup = function(platform, opts, cb) {
   case 'gentoo':
     cmd = 'chmod +x ' + scriptFile + '; rc-update add ' + p.basename(scriptFile) + ' default';
     break;
-    default :
+  case 'freebsd':
+    cmd = 'chmod +x ' + scriptFile;
+    break;
+  default :
     cmd = 'chmod +x ' + scriptFile + ' && update-rc.d ' + p.basename(scriptFile) + ' defaults';
     break;
   }
 
-  if (platform != 'darwin') {
+
+  if (platform == 'freebsd') {
+    cmd = 'su root -c "' + cmd + '"';
+  }else if (platform != 'darwin') {
     cmd = 'su -c "' + cmd + '"';
-  }else{
+  }else {
     cmd = 'pm2 dump';
   }
 
--- a/lib/scripts/pm2-freebsd.sh
+++ b/lib/scripts/pm2-freebsd.sh
@@ -9,7 +9,7 @@ load_rc_config $name
  
 : ${pm2_user="%USER%"}
  
-command="/usr/local/bin/${name}"
+command="%PM2_PATH%"
 pidfile="/home/${pm2_user}/.pm2/${name}.pid"
 start_cmd="${name}_start"
 stop_cmd="${name}_stop"
--- a/lib/scripts/pm2-freebsd.sh
+++ b/lib/scripts/pm2-freebsd.sh
@@ -7,7 +7,7 @@ rcvar=${name}_enable
  
 load_rc_config $name
  
-: ${pm2_user="root"}
+: ${pm2_user="%USER%"}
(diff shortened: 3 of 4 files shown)


soyuka:
Can you add this to the `pm2 startup` command?

https://github.com/jasin/PM2/blob/master/lib/CLI.js#L558

jasin:
I am not sure what is going on in your code to be fair. I would change it if I understood it a little better. Where is this idea of *_STARTUP_SCRIPT coming from? I do not find these files anywhere in the project. This is not making a whole lot of sense to me atm.

    if(!!~['systemd', 'centos', 'amazon', 'gentoo', 'darwin'].indexOf(platform)){
    script = cst[platform.toUpperCase() + '_STARTUP_SCRIPT'];
  }

the above code returns true when platform is in the list, so I add in freebsd to the list and then 
    script = FREEBSD_STARTUP_SCRIPT

now what? This is where I am lost, because there is no GENTOO_STARTUP_SCRIPT UBUNUTU* CENTOS* etc files. 

Guess I am just missing something, novice node/js programmer. enlighten me and I will take care of it. Thanks

soyuka:
Hi,

See [this commit](https://github.com/Unitech/PM2/commit/5cfd869bedfeb3574d6c07204378fe8267a1bc0f) as reference. Variable you're talking about are declared in [`constant.js`](https://github.com/Unitech/PM2/blob/master/constants.js#L42), so that their path can be easily replaced.

jasin:
Ok, well I completed the code changes. But I have some concerns. When running 

    sudo env PATH=$PATH:/user/local/bin/ pm2 startup freebsd -u jasin

it creates the file, everything works, but the pm2 daemon is started as root user because the command was run with sudo. In FreeBSD all rc scripts source rc.conf for variables that pertain to the script, this is where the user should be set. Not in the rc.script generated from -u option of pm2.
example rc.conf
    
    pm2_enable="YES"
    pm2_user="username"

Because pm2 daemon is started by root user in the above code( ie. sudo) it requires the daemon to be restarted after rc.conf has been edited to contain the two lines if you want pm2 daemon run as a normal user. FreeBSD requires at least one line in rc.conf for rc.scripts and that is whether the script should be enabled on startup or not. This is also the defacto file for all daemon related configurations

So I am kind of confused on how to best handle this situation or if it really doesn't matter. 

Please advise on how this is handled in other startup scripts or how you would like it to be done. This is beyond my pay grade, lol

soyuka:
I appreciate the effort!

So, about the user what we know is:
1. A startup script is usually created by the root user so that it can be put in the correct path (eg: `/etc/init.d` for ubunut)
2. PM2 might run as another user on startup

The workaround for the other startup script is that we replace `%USER%` with the cli option `-u myuser` ([for example](https://github.com/Unitech/PM2/blob/master/lib/scripts/pm2-init.sh#L21): `USER=%USER%`). In javascript it's done [here](https://github.com/Unitech/PM2/blob/master/lib/CLI.js#L586).

Then, we have a `super` function that will run:

```
    su - $USER -c "PATH=$PATH; PM2_HOME=$PM2_HOME $*"
```
so that the `super pm2 resurrect` is started with the correct user. `su - myuser -c whoami` will output `myuser` but was run as super user.

To your `pm2.freebsd` you only have to add something like:
```
PM2=%PM2_PATH% #binary path
USER=%USER% #user that'll run pm2

export PM2_HOME="%HOME_PATH%" #pm2 home (usually ~/.pm2 but can be different!)
```

and then use those variables as you already did :)!

Hope that's clear, please don't hesitate to commit code so that I can have a look or comment ;).

jasin:
No it doesn't really clear it up. What I am confused upon is when you invoke the 

    sudo pm2 startup freebsd

command it doesn't source the pm2.freebsd file to see who it should run as. If I run the above command it starts PM2 before it even install the script and because it is run as root it starts the daemon as root. 

this is only a problem when you run the command above, if the server reboots my script works as desired, it sources rc.conf and starts the pm2 daemon as a normal user. 

I will commit my code and let you look. I understand what you are saying but I am having a hard time seeing how it works if you do not source the startup file before starting the daemon.

maybe this will help, from this I am gathering that pm2 is started before the script is installed. This is where my problem lies.

    jasin:pm2/lib/scripts$ sudo pm2 startup freebsd
    Password:
    [PM2] Spawning PM2 daemon
    [PM2] PM2 Successfully daemonized
    [PM2] Generating system init script in /etc/rc.d/pm2
    [PM2] Making script booting at startup...
    [PM2] -freebsd- Using the command:
          su root -c "chmod +x /etc/rc.d/pm2"

    [PM2] Done.
    jasin:pm2/lib/scripts$ ps aux | grep PM2
    jasin    1151   0.0  3.6 637404  36400  -  Ss   12:52PM   0:01.98 PM2 v0.14.2: God Daemon (node)
    root     1241   0.0  3.5 635356  35812  -  Ss    1:32PM   0:00.65 PM2 v0.14.2: God Daemon (node)
    jasin    1259   0.0  0.2  18804   2224  0  S+    1:33PM   0:00.00 grep PM2

as you can see I now have two pm2 daemons running, the one started at boot time by user jasin and the one started by the above command as root.

jasin:
Ok man, I committed my changes. I'm still confused on first start why it thinks it needs to run as root, but it does. 

But, on reboot. With my code changes to CLI.js and constants.js  plus the pm2-freebsd.sh script and the two lines of code that are required by FreeBSD in rc.conf. this works for me on my server. Should work for others as well.

soyuka:
```
jasin:pm2/lib/scripts$ sudo pm2 startup freebsd
jasin:pm2/lib/scripts$ ps aux | grep PM2
root     1241   0.0  3.5 635356  35812  -  Ss    1:32PM   0:00.65 PM2 v0.14.2: God Daemon (node)
```
This is, either a bug, either a feature (depends on the point of view). 
When executing `pm2 -v` with no daemon instantiated yet (no pid), pm2 will daemonize itself and setup the configuration for the current user. To avoid it you could run `pm2 startup freebsd -u something && pm2 kill` if you're not using pm2 as root. 
It works like this to enable pm2 for several users at the same time, and so that one pm2 instance belongs to a single system user (security). 

Maybe we could fix this if command is `startup` and that there are no processes running.

jasin:
No update-rc.d, it just works out of the box. this is what rc.conf is for, you enable and disable daemons in there. Daemons are started as services and example for pm2 is

    sudo service pm2 start|stop|reload|status

jasin:
Ok, I have no idea why the build failed. All I did was update pm2-freebsd.sh to use %PM2_PATH% as requested. I'm done, works for me here and that was all that I cared about. 

Be my guess to fix whatever the hell is wrong. I'm out

soyuka:
Everything looks nice from here, thanks!

PR #1456, opened 2015-07-22
Force the su command to use bash

### LINKED ISSUE #890: CentOS startup scripts break with non-bash default shells
You end up with something like this:

```
/etc/init.d/pm2-init.sh status
Status for pm2:
PATH=(..your current $PATH here): Command not found.
┌──────────┬────┬──────┬───────┬────────┬───────────┬────────┬─────────────┬──────────┐
│ App name │ id │ mode │ PID   │ status │ restarted │ uptime │      memory │ watching │
├──────────┼────┼──────┼───────┼────────┼───────────┼────────┼─────────────┼──────────┤
```

For some of the servers I work on csh is the default shell, where "PATH=$PATH" is invalid syntax.
That path error can get pretty annoying and cause other people to freak out when they see errors.

One fix would be to add -s /bin/bash to the su line. That would probably be acceptable since pm2 is already relying on /bin/bash in the shebang line of the script itself.
I don't know how portable -s is on su though.

Gillingham:
Removing the shell detection because it doesn't really help, if the shell is (t)csh the startup command errors out due to syntax.
Closes #890

### DIFF
--- a/lib/scripts/pm2-init-centos.sh
+++ b/lib/scripts/pm2-init-centos.sh
@@ -29,7 +29,7 @@ export PM2_HOME="%HOME_PATH%"
 lockfile="/var/lock/subsys/pm2-init.sh"
 
 super() {
-    su - $USER -c "PATH=$PATH; PM2_HOME=$PM2_HOME $*"
+    su - $USER -s /bin/bash -c "PATH=$PATH; PM2_HOME=$PM2_HOME $*"
 }
 
 start() {
--- a/lib/scripts/pm2-init.sh
+++ b/lib/scripts/pm2-init.sh
@@ -23,20 +23,8 @@ USER=%USER%
 export PATH=%NODE_PATH%:$PATH
 export PM2_HOME="%HOME_PATH%"
 
-get_user_shell() {
-    local shell=$(getent passwd ${1:-`whoami`} | cut -d: -f7 | sed -e 's/[[:space:]]*$//')
-
-    if [[ $shell == *"/sbin/nologin" ]] || [[ $shell == "/bin/false" ]] || [[ -z "$shell" ]];
-    then
-      shell="/bin/bash"
-    fi
-
-    echo "$shell"
-}
-
 super() {
-    local shell=$(get_user_shell $USER)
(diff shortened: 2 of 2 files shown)


soyuka:
https://github.com/Unitech/PM2/pull/1141#issuecomment-91248481

Implementing the same function to the centos script seems to be a better idea. Did you test on centos only?

Gillingham:
Like I said, that script doesn't actually solve the problem. The issue I reported is when you're shell is something like csh or tcsh, not nologin or false.

If you're shell is not compatible with the bash syntax PM2 assumes exists then you get errors when running the su command.

By always forcing su to use bash we solve that issue, as well as the issue people seem to be having with nologin and false shells.

soyuka:
I agree but sometimes your environnement variables are linked to the shell you are working with. For example if I use `zsh`, I declare my environnement variable in `~/.zshrc`, it won't be sourced by `bash`. I know this can be seen as a bad practice but this might be an issue source.

@adamhadani any thoughts about this?

adamhadani:
my two cents -
Shell detection was added after initially I had a PR attempting go simply default to bash as suggested here - however this quickly became an issue as raised by another user which indeed relied on his .zshrc.
Perhaps we can modify shell detection logic a bit to default to /bin/bash in some specific cases?

soyuka:
> If your shell is not compatible with the bash syntax PM2 assumes exists then you get errors when running the su command.

Maybe that `PATH=PATH=$PATH:/.../.nvm/versions/node/v0.12.4/bin; PM2_HOME=/.../.pm2/ pm2 start` is not valid in some shells. 

I'll dig a bit more on this to try finding a better solution. We still have this issue about profile files if we use bash. 

Thanks for your two cents @adamhadani it's always helpful!

Gillingham:
> Maybe that PATH=PATH=$PATH:/.../.nvm/versions/node/v0.12.4/bin; PM2_HOME=/.../.pm2/ pm2 start is not valid in some shells.

That is what I've been trying to say, that's a bashism that does not carry over to csh type shells.

slavaGanzin:
```ps -p $$``` ?

Issue #1459, opened 2015-07-24
Startup for systemd fails if user is not root


bragma:
Hi,
I'm experiencing some problems with the "startup" feature for systemd. It fails to start the service correcty when pm2 is needs to be executed from a user different than root. I think I can explain the problem as a missing .pm2 folder and missing dump for the user. Please follow PR https://github.com/Unitech/PM2/pull/333 as a reference of what happens.

The command it executes is:

su -c "pm2 dump && pm2 kill && systemctl daemon-reload && systemctl enable pm2 && systemctl start pm2"

In a brief, the command to create the systemd script needs to be executed as root, since it must write to /etc/systemd folder. The problem is that when it executes "pm2 dump", it dumps in root's home, creating "/root/.pm2/dump.pm"

But later, when it tries to start the service with "systemctl start pm2", the service is started as user "user", which is missing the dump in its own folder (ex: /home/user/.pm2/dump.pm). 

```
# pm2 startup systemd
[PM2] Spawning PM2 daemon
[PM2] PM2 Successfully daemonized
[PM2] You have to run this command as root. Execute the following command:
      sudo env PATH=$PATH:/usr/bin pm2 startup systemd -u user
# sudo env PATH=$PATH:/usr/bin pm2 startup systemd -u user
[PM2] Generating system init script in /etc/systemd/system/pm2.service
[PM2] Making script booting at startup...
[PM2] -systemd- Using the command:
      su -c "pm2 dump && pm2 kill && systemctl daemon-reload && systemctl enable pm2 && systemctl start pm2"
Command failed: [PM2][WARN] No process found
Job for pm2.service failed. See "systemctl status pm2.service" and "journalctl -xe" for details.
# sudo systemctl status pm2
● pm2.service - PM2 next gen process manager for Node.js
   Loaded: loaded (/etc/systemd/system/pm2.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Fri 2015-07-24 13:23:08 UTC; 58s ago
  Process: 2407 ExecStart=/usr/local/lib/node_modules/pm2/bin/pm2 resurrect (code=exited, status=1/FAILURE)
 Main PID: 2019 (code=exited, status=0/SUCCESS)

Jul 24 13:23:07 box systemd[1]: Starting PM2 next gen process manager for Node.js...
Jul 24 13:23:08 box pm2[2407]: [PM2] Resurrecting
[...]

PR #1463, opened 2015-07-24
Fix for startup systemd not working when user is not root

bragma:
Fix for https://github.com/Unitech/PM2/issues/1459


soyuka:
Thanks, lgtm from here! Just waiting for tests to pass.

Unitech:
Thanks!

Test cannot pass I made a change in the module system breaking some test suits, it's fixed in dev

PR #1483, opened 2015-07-29
fix #1415 add logrotate command

### LINKED ISSUE #1415: Logrotate integration
Hi there,

Recently I've experienced an edge case with PM2 log files. My server's storage was full because of my application's PM2 log file -which was 25GB in size :fearful: As a result, the application stopped working. I think this is something serious that should be considered when setting up default behavior with collecting log files.

I guess integration with logrotate can be an option. Just like nginx. As you probably know, nginx -by default- uses logrotate to compress and remove old log files. IMHO PM2 should have something similar which would prevent edge cases like I experienced. 

Thanks!

soyuka:


soyuka:
As `/etc/logrotate.d` will only be available with an elevated user I think this is the best solution. Feel free to merge as soon as tests are good to go :).

Issue #1256, opened 2015-05-12
Shouldn't default exec_mode be fork?


DomiR:
https://github.com/Unitech/PM2/blob/master/lib/Common.js#L117

Default mode should probably be fork_mode: 
```
  if(typeof app.instances == 'undefined'){
    app.instances = 1;
  }

  ...

  /**
   * Here we put the default exec mode
   */
  if (!app.exec_mode && app.instances > 0) {     // check for app.instances > 1 instead
    app.exec_mode = 'cluster_mode';
  } else if (!app.exec_mode) {
    app.exec_mode = 'fork_mode';
  }
```

jorge-d:
The default mode is `fork_mode`, unless you specify the `-i` option that will run the cluster mode. I don't see anything unusual here?

DomiR:
Well I'm not sure if the depicted prepareAppConf is used by pm2, but it is in a third party management tool([pod](https://github.com/yyx990803/pod)). And it seems to me, unless you set instances explicitly to zero the default exec mode would be `cluster_mode`.

jorge-d:
No, the default mode is `fork`:
```shell
$ pm2 start app.js
[PM2] Spawning PM2 daemon
[PM2] PM2 Successfully daemonized
[PM2] Process app.js launched
┌──────────┬────┬──────┬──────┬────────┬─────────┬────────┬─────────────┬──────────┐
│ App name │ id │ mode │ pid  │ status │ restart │ uptime │ memory      │ watching │
├──────────┼────┼──────┼──────┼────────┼─────────┼────────┼─────────────┼──────────┤
│ app      │ 0  │ fork │ 6572 │ online │ 0       │ 0s     │ 17.945 MB   │ disabled │
└──────────┴────┴──────┴──────┴────────┴─────────┴────────┴─────────────┴──────────┘
 Use `pm2 show <id|name>` to get more details about an app

```

DomiR:
Yes, I know. But the CLI you're showing is not using the Common.prepareAppConf for whatever reason. 

See https://github.com/Unitech/PM2/blob/master/lib/CLI.js#L93.
```
var appConf = {};

  if (!!opts.executeCommand)
    conf.exec_mode = 'fork';
  else if (opts.instances !== undefined)
    conf.exec_mode = 'cluster';
  else
    conf.exec_mode = 'fork';
```
[...]

PR #1492, opened 2015-08-03
Change default exec mode to fork in prepareConf

DomiR:
Should behave the same as cli (cli should probably use common.prepareConf instead of verifying the conf manually).

### DIFF
--- a/lib/Common.js
+++ b/lib/Common.js
@@ -112,7 +112,7 @@ Common.prepareAppConf = function(app, cwd, outputter) {
   /**
    * Here we put the default exec mode
    */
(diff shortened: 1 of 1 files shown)


jshkurti:
Some people use cluster mode with only 1 instance for various reasons and it doesn't hurt anybody so far :P 
Why wouldn't we allow such usage ?

DomiR:
Because of this: https://github.com/Unitech/PM2/issues/1256

DomiR:
Why does the cli default to fork then?

jshkurti:
Ok I got it. Thanks for the PR :)

PR #1523, opened 2015-08-13
ADVANCED_README.md: add "min_uptime" description, fix #1331

### LINKED ISSUE #1331: `min_uptime` missing from JSON app declaration doc?
According to the [ADVANCED_README.md#option-1](https://github.com/Unitech/PM2/blob/master/ADVANCED_README.md#options-1), there is an option called `min_uptime` that is not referenced into [ADVANCED_README.md#list-of-all-json-declaration-fields-avaibles](https://github.com/Unitech/PM2/blob/master/ADVANCED_README.md#list-of-all-json-declaration-fields-avaibles).

I first thought this option was deprecated but I saw in many issues that many people still use it. That should be just a miss in the documentation.

aronwoost:


aronwoost:
Not sure why test is failing. Sorry for not PR'ing to development, should be fine in this case since it's only docs. Thanks for PM2!

jshkurti:
Thanks :)

PR #1529, opened 2015-08-14
Add nvm compatibility to PM2 systemd startup

dguenther:
I was running into some issues when running `pm2 systemd startup` on a system with node installed via nvm.

Initially, the environment is set correctly through `sudo su -c 'env PATH=$PATH:/usr/bin pm2 startup systemd -u xxx' `, but once I ran that command, the script it built:

`su web -c "pm2 dump && pm2 kill" && su root -c "systemctl daemon-reload && systemctl enable pm2 && systemctl start pm2"`

Doesn't run `pm2` with a valid path to node. So I added the `env` command to each of the `pm2` calls. That caused the script to execute, but the systemd unit also didn't have the path set, so I had to run each command in a shell so I could add the `env` command.

I may be going about this wrong, and if so I apologize. It does mean I'll have to regenerate the startup script whenever I upgrade my version of node. Still, hopefully this is useful to anyone else trying to do something similar.

### DIFF
--- a/lib/CLI.js
+++ b/lib/CLI.js
@@ -615,14 +615,16 @@ CLI.startup = function(platform, opts, cb) {
 
   var cmd;
   var cmdAsUser;
+  var cmdPrefix;
 
   printOut(cst.PREFIX_MSG + 'Making script booting at startup...');
 
   switch (platform) {
   case 'systemd':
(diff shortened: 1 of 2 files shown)


dguenther:
Closing and reopening against development.

PR #1530, opened 2015-08-14
Add nvm compatibility to PM2 systemd startup

dguenther:
I was running into some issues when running `pm2 startup systemd` on a system with node installed via nvm.

Initially, the environment is set correctly through `sudo su -c 'env PATH=$PATH:/usr/bin pm2 startup systemd -u xxx'`, but once I ran that command, the script it built:

`su web -c "pm2 dump && pm2 kill" && su root -c "systemctl daemon-reload && systemctl enable pm2 && systemctl start pm2"`

Doesn't run pm2 with a valid path to node. So I added the `env` command to each of the `pm2` calls. That caused the script to execute, but the systemd unit also didn't have the path set, so I had to run each command in a shell so I could add the `env` command.

I may be going about this wrong, and if so I apologize. It does mean I'll have to regenerate the startup script whenever I upgrade my version of node. Still, hopefully this is useful to anyone else trying to do something similar.

### DIFF
--- a/lib/CLI.js
+++ b/lib/CLI.js
@@ -615,14 +615,16 @@ CLI.startup = function(platform, opts, cb) {
 
   var cmd;
   var cmdAsUser;
+  var cmdPrefix;
 
   printOut(cst.PREFIX_MSG + 'Making script booting at startup...');
 
   switch (platform) {
   case 'systemd':
+    cmdPrefix = 'env PATH=$PATH:' + p.dirname(process.execPath);
     cmdAsUser = [
-      'pm2 dump', //We need an empty dump so that the first resurrect works correctly
-      'pm2 kill',
+      cmdPrefix + ' pm2 dump', //We need an empty dump so that the first resurrect works correctly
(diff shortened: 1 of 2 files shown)


animedbz16:
Not at my computer, but cant this be solved by adding a nvm use vX.Y into
your profile as the init.d shouldnt need to know what versiob of node to
start for your user?

dguenther:
It's definitely possible that this isn't the right fix. Embedding the node path in the service is not ideal either. As far as I can tell, adding `nvm use` to the profile would work if the `sudo su user -c` command were run as a login shell (so `sudo su -l user -c`), and the systemd unit used `runuser` instead of embedding the node path. That might be a better change to make, if one is needed.

Another option is to assume nvm users will install a node version globally, which would avoid this issue.

Unitech:
All of this is fixed with the new systemd support on PM2 2.2.x

```
$ npm install pm2@latest -g
$ pm2 startup systemd
```

PR #1531, opened 2015-08-14
Warn pm2 startup users that might use systemd

### LINKED ISSUE #1319: Problems with `pm2 startup` on ubuntu 15.04
Looks like on Ubuntu 15.04 (first Ubuntu version running systemd instead of upstart by default), you have to do `pm2 startup systemd` explicitly.

In my case, just `pm2 startup` did not seem to make my processes start at startup.

dguenther:
Currently, users of distributions like Ubuntu 15.04 and CentOS/RHEL 7 have to know that their distribution is running `systemd` in order to set up `pm2 startup` correctly. We can check the version listed in the `/etc/os-releases` file to see if it's likely that a user is on one of these systems. If so, this branch outputs a warning:

```
[PM2][WARN] Your distribution (Ubuntu 15.04) uses systemd by default. Unless you've changed this, you should run "pm2 startup systemd" instead.
```

Builds off work in #1120. Resolves #1319

### DIFF
--- a/lib/CLI.js
+++ b/lib/CLI.js
@@ -11,6 +11,7 @@ var util                 = require('util');
 var vm                   = require('vm');
 var chalk                = require('chalk');
 var exec                 = require('child_process').exec;
+var ini                  = require('ini');
 
 var p                    = path;
 
@@ -562,6 +563,25 @@ CLI.actionFromJson = function(action, file, jsonVia, cb) {
  * @param {string} platform type (centos|redhat|amazon|gentoo|systemd)
  */
 CLI.startup = function(platform, opts, cb) {
+  if (platform == 'linux' || platform == 'redhat' || platform == 'ubuntu' || platform == 'centos') {
+    try {
+      var osRelease = ini.parse(fs.readFileSync('/etc/os-release', 'utf-8'));
+      var mightUseSystemd = false;
+      switch(osRelease.ID) {
+      case 'ubuntu':
+        mightUseSystemd = osRelease.VERSION_ID >= 15.04;
+        break;
+      case 'centos':
+      case 'rhel':
+        mightUseSystemd = osRelease.VERSION_ID >= 7;
+        break;
+      }
+      if (mightUseSystemd) {
+        console.log(cst.PREFIX_MSG_WARNING + 'Your distribution (%s) uses systemd by default. Unless you\'ve changed this, you should run "pm2 startup systemd" instead.', osRelease.PRETTY_NAME);
+      }
+    } catch (e) { }
+  }
+
   if (process.getuid() != 0) {
     return exec('whoami', function(err, stdout, stderr) {
       console.error(cst.PREFIX_MSG + 'You have to run this command as root. Execute the following command:\n' +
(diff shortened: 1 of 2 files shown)


Unitech:
@soyuka @alavit-d thoughts?

soyuka:
@dguenther what about #1120, could you give an opinion about it? Should me consider merging it/reviewing it?

Thanks!

dguenther:
Yup, it looks like `existsSync` was deprecated, so I'll change it to use `accessSync` instead.

#1120 makes good changes around running processes as the correct user, but I have a few concerns with how it determines the platform. For one thing, a user on CentOS 7 might specify `centos` and expect PM2 to assume systemd, but in that pull request, PM2 will assume sysv. Even if the branch included `centos` as well as `linux` when checking for systemd, I'm concerned that if users have overridden systemd as the default init system, the `/etc/os-releases` file might still exist. Then we'd need a way for users to specify sysv as their init. Finally, the branch only addresses centos, not ubuntu and redhat.

I looked into methods for detecting the init system, but it seems almost impossible to do reliably, so that's why I think it's better to log a warning instead.

soyuka:
> Yup, it looks like existsSync was deprecated, so I'll change it to use accessSync instead.

Oh gosh, I was not aware of this! Indeed, `accessSync` seems appropriate.

> I looked into methods for detecting the init system, but it seems almost impossible to do reliably, so that's why I think it's better to log a warning instead.

Agreed.

It's such a mess with all those startup mechanisms, hope that we'll get to a nice and portable solution one day -_-.

dguenther:
Hmm, `accessSync` and `statSync` both also throw ENOENT if the file does not exist. I'm not sure we can avoid the `try/catch` on this, which does make me curious why `existsSync` was deprecated...

soyuka:
Use the `try/catch` thing, I have no alternatives :/

dguenther:
All right, I'll leave it as-is. Would you like me to fix up anything else on this branch before considering it for merge?

Tjatse:
> The file `/etc/os-release` takes precedence over `/usr/lib/os-release`. Applications should check for the former, and exclusively use its data if it exists, and only fall back to `/usr/lib/os-release` if it is missing. Applications should not read data from both files at the same time. `/usr/lib/os-release` is the recommended place to store OS release information as part of vendor trees. `/etc/os-release` should be a relative symlink to `/usr/lib/os-release`, to provide compatibility with applications only looking at `/etc`. A relative symlink instead of an absolute symlink is necessary to avoid breaking the link in a chroot or initrd environment such as dracut.

dguenther:
I assume that comment means to check for the `os-release` file in both locations? I've added that to my branch. In practice, I'm not sure that distros have established any consistency on this. Ubuntu 15.04 doesn't have a file at `/usr/lib/os-release`, for example.

slavaGanzin:
Ah? Why are you parsing something if you can just ```exec('systemctl', (ThisWouldBeAnErrorIfNoSystemd) =>)```

Unitech:
Now on PM2 2.2.x, there is an automatic detection of the running init system.

```
$ npm install pm2 -g
$ pm2 startup
```

PR #1580, opened 2015-09-03
Fix #1578 + test ES6 with nodejs 

### LINKED ISSUE #1578: pm2 API with watch string or array crash on windows + workaround
I've found an error and a workaround when using pm2 API.

Problem is on watcher (line 47).
It appears that process.env.PWD is ``undefined`` on windows.

Find below the code I'm using and the fix by assigning ``process.env.PWD`` with ``process.cwd()`` value.

````
var gulp         = require('gulp');
var pm2 = require('pm2');

gulp.task('pm2', function(done) {
  var conf = require('./config/pm2/base.json');

  // bug on pm2 (watcher - 47) which expects process.env.PWD to exist
  process.env.PWD = process.cwd();

  pm2.connect(true, function() {
    pm2.start(conf, function(err /*, apps*/) {
      if (err) throw err;
      console.log('pm2 started');
      pm2.streamLogs('all', 0);
      done();
      // pm2.disconnect();
    });
  });
});

````

Problem identified is in file ``lib/Watcher.js`` on the following line

````
...
      watch = [].concat(pm2_env.watch).map(function(dir) {
          return p.resolve(process.env.PWD, dir);
      });
...
````

soyuka:
- Don't exit es6 tests if nodejs is detected
- Fix #1578 by using pm_cwd instead of env.PWD

### DIFF
--- a/bin/pm2
+++ b/bin/pm2
@@ -331,7 +331,6 @@ commander.command('updatePM2')
   .action(function() {
     CLI.updatePM2();
   });
-
 commander.command('update')
   .description('(alias) update in-memory PM2 with local PM2')
   .action(function() {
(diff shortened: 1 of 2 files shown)


Unitech:
:+1: :)

PR #1589, opened 2015-09-06
Fix install specific version

soyuka:
Like: `pm2 install module@2.1.0-beta`
Remove `force` option from rm, not needed and dangerous


Unitech:
thanks, merged in dev

PR #1597, opened 2015-09-09
Update travis Node.js version & remove io.js

demohi:


soyuka:
We will not remove io.js right now.

I'm merging #1598 instead

PR #1604, opened 2015-09-10
Upating watcher.js to work with windows file systems

JWoodJr:
On windows file systems not setting the type of the arguments to path.resolve causes an error that breaks Watcher, manually setting the types to string fixes this.


soyuka:
Fixed here: https://github.com/Unitech/PM2/commit/f6847bae3b7711ba573e6e210e378c4fcd7ad05f
not released yet!

To try it out: `npm i -g Unitech/PM2#development`

PR #1673, opened 2015-10-06
Be sure to call the callback passed to `pm2.deploy`

### LINKED ISSUE #1671: [Deploy] pm2.deploy never invokes its callback on success

ide:
Sometimes the callback that is passed to pm2.deploy is never called. This means tools that use PM2 cannot know when it has completed deploying the code. This commit makes sure we call the callback before exiting.

Fix #1671


Unitech:
awesome thanks :+1:

PR #1705, opened 2015-10-20
Allows to install packages from tarball generated with 'npm pack'

flozz:
Hello,

This PR enables PM2 to install packages from tarball (generated with the `npm pack` command), if the tarball is properly named:

    pm2 install myapp.tgz
    pm2 install myapp-1.0.0.tgz
    pm2 install http://example.com/packages/myapp-1.0.0.tgz

PR #1713, opened 2015-10-23
Allows to install packages from tarball generated with 'npm pack'  #2

flozz:
Hello,

This PR enables PM2 to install packages from tarball (generated with the `npm pack` command), if the tarball is properly named:

    pm2 install myapp.tgz
    pm2 install myapp-1.0.0.tgz
    pm2 install http://example.com/packages/myapp-1.0.0.tgz

// PR #1705 but, on the right branch :)


Unitech:
thanks!

Issue #1720, opened 2015-10-27
Wrong location during pm2 dump on 'startup systemd -u USER'


pszypowicz:
pm2 version: 0.15.7
node version: 4.2.1
npm version: 2.14.7
system centos 7 x86_64

While using 
```bash
pm2 startup systemd -u node
```
i receive information: 

```
$ pm2 startup systemd -u node
[PM2] Spawning PM2 daemon
[PM2] PM2 Successfully daemonized
[PM2] Generating system init script in /etc/systemd/system/pm2.service
[PM2] Making script booting at startup...
[PM2] -systemd- Using the command:
      su node -c "pm2 dump && pm2 kill" && su root -c "systemctl daemon-reload && systemctl enable pm2 && systemctl start pm2"
Command failed: /bin/sh -c su node -c "pm2 dump && pm2 kill" && su root -c "systemctl daemon-reload && systemctl enable pm2 && systemctl start pm2"
internal/child_process.js:298
    throw errnoException(err, 'spawn');
    ^

Error: spawn EACCES
    at exports._errnoException (util.js:874:11)
    at ChildProcess.spawn (internal/child_process.js:298:11)
    at Object.exports.spawn (child_process.js:339:9)
    at Object.launchDaemon (/usr/lib/node_modules/pm2/lib/Satan.js:351:40)
    at /usr/lib/node_modules/pm2/lib/Satan.js:53:20
    at /usr/lib/node_modules/pm2/lib/Satan.js:399:14
    at doNTCallback0 (node.js:417:9)
    at process._tickDomainCallback (node.js:387:13)

----- Are you sure you use the right platform command line option ? centos / redhat, amazon, ubuntu, gentoo, systemd or darwin?
```

I see issue it this line: 
```
/bin/sh -c su node -c "pm2 dump && pm2 kill" 
```
which leaves HOME for commands execution same as root. Effectively he is trying to write to root HOME and surprised it cannot :) I will follow with pull request that basically will use **su -l** for those first command. As I observed, once .pm2 folder is created in user home dir, its using this, not causing EACCES errors.

soyuka:
https://github.com/Unitech/pm2/issues/1654#issuecomment-153181749 `--hp home/path` should fix your issue. @Unitech thoughts?

Unitech:
The -l option can fix this issue, but with the new `--hp home/path` option (shown when using pm2 startup) should fix this issue, please upgrade your PM2

PR #1721, opened 2015-10-27
Fix creating pm2 files in correct user $HOME folder

pszypowicz:
https://github.com/Unitech/pm2/issues/1720

### DIFF
--- a/lib/CLI.js
+++ b/lib/CLI.js
@@ -682,7 +682,7 @@ CLI.startup = function(platform, opts, cb) {
   }
 
   if (platform == 'systemd') {
-    cmd = 'su ' + user + ' -c "' + cmdAsUser + '" && su root -c "' + cmd + '"';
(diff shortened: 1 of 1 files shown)


Unitech:
what does the -l option?

Unitech:
https://github.com/Unitech/pm2/commit/419b748e58dc7874d6c5f25723032aa584c7c9ad

pszypowicz:
After the man pages:

```
       -, -l, --login
              Start the shell as a login shell with an environment similar
              to a real login:

                 o      clears all the environment variables except TERM

                 o      initializes the environment variables HOME, SHELL,
                        USER, LOGNAME, and PATH

                 o      changes to the target user's home directory

                 o      sets argv[0] of the shell to '-' in order to make
                        the shell a login shell
```
which here will simply change home directory to user we selected in '-u' option. I think that allowing it to be ANY home directory for pm2 dump won't work as expected.

soyuka:
closing in favor of #1902

PR #1726, opened 2015-10-29
Fix link attached to travis-ci build badge

Jonnymcc:


Unitech:
previously fixed thanks

PR #1745, opened 2015-11-04
Watcher to log the path which triggered restart

tiblu:
Watcher to log the path which triggered restart which greatly improves debugging infinite restart loops.


tiblu:
Your commit changes the original log message whilst I kept it backward compatible. Is that a good idea?

PR #1751, opened 2015-11-06
fix syntax error

NinoFocus:


Unitech:
this is the behavior expected

PR #1770, opened 2015-11-17
Fix all misspellings of "unknown"

rtheunissen:
"Unknow" should be "Unknow**n**"


rtheunissen:
Should have compared against development. :boom:

Issue #1781, opened 2015-11-20
PM2 home directory on Windows


hmny:
I have a `logs` folder in `C:\Users\username\.pm2\logs`, but whenever I try to start `pm2` I am getting this error in my Logs:
```
PM2: 2015-11-20 13:23:32: Starting execution sequence in -fork mode- for app name:app id:0
PM2: 2015-11-20 13:23:32: Trace: [Error: can not create directories (logs/pids):ENOENT: no such file or directory, mkdir 'D:\Users\username\.pm2\logs']
```
So I think this is a config issue. How can I change the PM2 home directory to `C:\Users\username\.pm2` instead of `D:\Users\username\.pm2`?

Unitech:
What are the values of HOME and HOMEDRIVE environment variables?

https://github.com/Unitech/pm2/blob/master/constants.js#L16

hmny:
Well none of them were pointed to `D:\`. Anyway, I solved it by reinstalling pm2.

vmarchaud:
If anyone still have this issue, you can set a PM2_HOME in your env to change the home directory of pm2.

LarsVoith:
> If anyone still have this issue, you can set a PM2_HOME in your env to change the home directory of pm2.

also try echo %PM2_HOME% to test, I could not see the system variables until I rebooted in my project root.

PR #1843, opened 2015-12-17
Fixing home directory on windows #1781

standy:
I got the issue on windows, HOME was relative to drive. `process.env.HOME = "\Users\user"` 
Resolving HOME by HOMEDRIVE will no affect if HOME is absolute, and will fix the issue if HOME is relative


Unitech:
Thanks for the fix, but I had to edit some stuff

PR #1850, opened 2015-12-20
load configuration from /etc/default/pm2

### LINKED ISSUE #1844: PM2-init.sh - load defaults from /etc/default/pm2
The pm2-init.sh script should load a script from /etc/default/pm2 that will be used to override its defaults (That way you don't have to modify pm2-init.sh manually.

### LINKED ISSUE #1845: add ulimit -n override in pm2-init.sh
using an ENV variable like here for example: MAX_OPEN_FILES

shaharmor:
fixes #1844 
fixes #1845

### DIFF
--- a/lib/scripts/pm2-init.sh
+++ b/lib/scripts/pm2-init.sh
@@ -19,9 +19,25 @@
 NAME=pm2
 PM2=%PM2_PATH%
 USER=%USER%
+DEFAULT=/etc/default/$NAME
 
 export PATH=%NODE_PATH%:$PATH
 export PM2_HOME="%HOME_PATH%"
+
+# The following variables can be overwritten in $DEFAULT
+
+# maximum number of open files
+MAX_OPEN_FILES=
+
+# overwrite settings from default file
+if [ -f "$DEFAULT" ]; then
+	  . "$DEFAULT"
+fi
+
+# set maximum open files if set
+if [ -n "$MAX_OPEN_FILES" ]; then
+    ulimit -n $MAX_OPEN_FILES
+fi
 
 get_user_shell() {
     local shell=$(getent passwd ${1:-`whoami`} | cut -d: -f7 | sed -e 's/[[:space:]]*$//')


Unitech:
Looks great, thank you!

Unitech:
I did some test, unfortunatelly environment variables loaded from /etc/default/pm2 are not passed to the pm2 command :/

because of the `su -` command: https://github.com/Unitech/pm2/blob/master/lib/scripts/pm2-init.sh#L55

I'm trying some stuff, if you have any idea it's welcome

shaharmor:
How did you test it? i am using something similar in production now

Unitech:
Replace:

```
su - $USER -s $shell -c "PATH=$PATH; PM2_HOME=$PM2_HOME $*"
```

By:

```
su - $USER -s $shell -c "env; PATH=$PATH; PM2_HOME=$PM2_HOME $*"
```

I've added PM2_KILL_TIMEOUT in /etc/default/pm2 and the variable is not present when doing the `env` command

Unitech:
I found the solution, when importing variables we must do [`set -a`](https://gist.github.com/Unitech/4c736f2076e4f37902ce#file-gistfile1-txt-L24)  then when we finished we do [`set -a`](https://gist.github.com/Unitech/4c736f2076e4f37902ce#file-gistfile1-txt-L46)

On the su command we just need to pass the [-m option](https://gist.github.com/Unitech/4c736f2076e4f37902ce#file-gistfile1-txt-L61) to preserve environment variable

But now the problem is that if someone use the /etc/default/pm2 file for configuration variables, PM2 is not aware itself of this file, so I will continue to think about this way of configuring PM2

shaharmor:
I'm not sure this is supposed to be used to set environment variables, but instead just bash variables

shaharmor:
Its not supposed to replace PM2 environment variables (Which for themselves should move to the app json), but just startup variables like ulimit

Unitech:
Yes I was thinking that it could be a good way to configure some internals PM2 variables but yes you are right, let's keep the /etc/default/pm2 for startup variables

PR #1851, opened 2015-12-22
Deploy to default environment.

ivasilov:
When the user deploys without supplying an argument for environment, if there's only 1 defined environment in the ecosystem.json, deploy to that environment.

This is a followup to #1774.


soyuka:
thanks!

PR #1902, opened 2016-01-19
Rework startup command code

### LINKED ISSUE #1888: Wrong user home set when generating pm2-init.sh
Just followed the instructions in http://pm2.keymetrics.io/docs/usage/startup/#user-permissions and ran this command `$ pm2 startup ubuntu -u node`

Then had a look at `/etc/init.d/pm2-init.sh` here are the first top lines of that fine:

```
NAME=pm2
PM2=/usr/lib/node_modules/pm2/bin/pm2
USER=nodejs
DEFAULT=/etc/default/$NAME

export PATH=/usr/bin:$PATH
export PM2_HOME="/home/michael.heuberger/.pm2"
```

`USER` is correct. But `PM2_HOME` is wrong. Should be `/home/nodejs/.pm2`

estliberitas:
This commit brings several things:
* Move command code to separate script
* Refactoring of code in order to be async and provide better error-handling
* Support for future integration with Windows OS
* Print & handle errors in a more clean way
* Detect user home directory

Fixes #1888

### DIFF
--- a/lib/CLI.js
+++ b/lib/CLI.js
@@ -28,6 +28,7 @@ var Modularizer          = require('./Modularizer.js');
 var Configuration        = require('../lib/Configuration.js');
 
 var CLIDeploy = require('./CLI/Deploy.js');
+var CLIStartup = require('./CLI/Startup.js');
 
 CLI.pm2Init = function() {
   if (!fs.existsSync(cst.PM2_ROOT_PATH)) {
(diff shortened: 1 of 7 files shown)


estliberitas:
Check this out, guys! I did not add tests yet. Probably gonna add them in next PR.

Btw, I know I almost re-wrote everything in startup command but original sequence of actions is preserved for sure. Please, review it. I've tested this on OS X and now playing with CentOS.

/cc @soyuka @Unitech @binarykitchen @JNK

estliberitas:
Also I fixed warning printed in case user does not have superuser privileges. Previously printed command was `sudo su -c "blablabla"`. Actually, this can't work on OS X, but having `sudo su root -c "blablabla"` seems to work everywhere.

jcollum:
Thanks guys. I'm using the `--hp` option and exporting PM2_HOME but still finding that the pm2 startup wants to use my home directory instead of the one specified. Commenting here so I'll be alerted when the merge is done (hopefully).

estliberitas:
@soyuka Well, I used `fs.stat`. Just to be compatible with oldie like 0.10 - I saw in Travis that it's still supported, correct?

soyuka:
Still available on the old commit: https://github.com/estliberitas/PM2/commit/d394bddbbe9622be3f72605d398bf89bcb93655e :)

> @soyuka Btw, what about moving printError into Common or Utility? As for me, it seems to be useful for error logging, when both message and error should be printed.

Indeed I almost said it but then I saw that printError was calling `Common.printError` I needed to think a bit further. 

This is my proposal:

Small refactor of the `Common.printError` function:

```
//Common.js
/**
 * Description
 * @method printError
 * @param {} msg
 * @return CallExpression
 */
Common.printError = function() {
  if (process.env.PM2_SILENT || process.env.PM2_PROGRAMMATIC === 'true') return false;

  var args = [].slice.call(arguments)  
  for(var i in args) {
    if (args[i] instanceof Error)
      console.error(args[i].stack);
    else
      console.error(args[i]);
  }
};
```

And now you can call `Common.printError` directly. I don't see why errors should not be logged to stderr (see `Common.printOut`).

estliberitas:
@soyuka Sounds reasonable. Let it be in other PR, maybe? So commits are consistent.

soyuka:
I made a PR on your branch, I think that we can handle the `Error` object only once, I kept specific messages because they are nice :).

estliberitas:
@soyuka I see you removed error from `Common.printError()`. I'm wondering how and do we display original error? Because in case of non-programmatic usage original error seems to be lost.

estliberitas:
@soyuka Btw, why not use `callback(err)` instead of `callback({msg: err})`. We always can add callback to generate function which would change `err` to `{msg: err}` if this is ever needed.

soyuka:
Where is the `callback({msg: err})`?

estliberitas:
@soyuka In code there at least 5 such cases of execution callback with error. :-)

estliberitas:
I was wrong - nine.

soyuka:
might be for the programmatic response, I'll have to investigate

estliberitas:
@soyuka Well if it's so... Then it's done in a quite non-node way.

The thing is when I saw it first, I thought this way is correct, so I made more such executions. And overall, in project, such way of error reporting is used as usual thing. So it's fine. But for one command, I think it's better to return callbacks with actual error or message and only in top-level function wrap it with `{msg: err.message}` or so. Please, check. We should have clean PR.

estliberitas:
@soyuka Rebased, so you can check the idea.

estliberitas:
This means you have to rebase your PR too.

estliberitas:
@soyuka Guess, we're ready to merge this one too.

estliberitas:
@soyuka Hold, please.

estliberitas:
@soyuka Now it's good to go. I cherry-picked your commit instead of merging. Now commits history is clean.

estliberitas:
@Unitech Mind adding me to collaborators? Or no opportunity ATM? :smile:

soyuka:
Mind doing this PR also on development? Then I can send this on travis to see how it goes :).
thanks!

Issue #1764, opened 2015-11-13
pm2 start and never return


bangbang93:
I use json declare my application 
```javascript
{
  "apps" : [{
    "name"        : "push",
    "script"      : "./index.js",
    "watch"       : true,
    "merge_logs"  : true,
    "env": {
      "DEBUG": "cofactories-push:*",
      "NODE_ENV": "production"
    }
  }]
}
```
then I use "pm2 start process.json" to start my application .But it just stop here 
![image](https://cloud.githubusercontent.com/assets/3430784/11138413/ec3898ac-89fc-11e5-9ad7-a10434eb1df9.png)
And pm2 list can't find it .
But if I use "pm2 start index.js", it works 
pm2 0.15.10
node 4.2.2

bangbang93:
when i use "pm2 restart push --watch",daemon even stop running(I use another ssh terminal run "pm2 list",got "spawing pm2 daemon")and lost my process list
```
2015-11-13 21:56:43: App name:www id:12 exited with code SIGINT
2015-11-13 21:56:43: Process with pid 8592 killed
2015-11-13 21:56:43: Starting execution sequence in -fork mode- for app name:www id:12
2015-11-13 21:56:43: App name:www id:12 online
2015-11-13 22:02:47: [PM2] Error caught by domain:
TypeError: God.watch._watchers[pm2_env.name].close is not a function
    at Object.ClusterMode.God.watch.enable (/home/cofactories/.nvm/versions/node/v4.2.2/lib/node_modules/pm2/lib/Watcher.js:20:41)
    at God.restartWatch (/home/cofactories/.nvm/versions/node/v4.2.2/lib/node_modules/pm2/lib/God/ActionMethods.js:576:19)
    at Server.onmessage (/home/cofactories/.nvm/versions/node/v4.2.2/lib/node_modules/pm2/node_modules/pm2-axon-rpc/lib/server.js:105:6)
    at emitTwo (events.js:87:13)
    at RepSocket.emit (events.js:172:7)
    at Parser.<anonymous> (/home/cofactories/.nvm/versions/node/v4.2.2/lib/node_modules/pm2/node_modules/pm2-axon/lib/sockets/rep.js:51:15)
    at emitOne (events.js:77:13)
    at Parser.emit (events.js:169:7)
    at Parser._write (/home/cofactories/.nvm/versions/node/v4.2.2/lib/node_modules/pm2/node_modules/pm2-axon/node_modules/amp/lib/stream.js:91:16)
[...]

PR #1909, opened 2016-01-22
Fix Watcher to be able to run application with whatever name

estliberitas:
Hi guys!

This is only a proposition of a fix for #1764, because first Automattic/cli-table#83 should be accepted and updated. Once they release fixed version, I'll rebase PR with `package.json` update also. And then, all tests will pass, and we'll be good to go. :wink: 

/cc @Unitech @soyuka


soyuka:
May you PR on development instead of master? thx.

estliberitas:
@soyuka np

PR #1910, opened 2016-01-22
Fix Watcher to be able to run application with whatever name

estliberitas:
Hi guys!

This is only a proposition of a fix for #1764, because first Automattic/cli-table#83 should be accepted and updated. Once they release fixed version, I'll rebase PR with `package.json` update also. And then, all tests will pass, and we'll be good to go. :wink: 

/cc @Unitech @soyuka

### DIFF
--- a/lib/Common.js
+++ b/lib/Common.js
@@ -51,10 +51,6 @@ Common.prepareAppConf = function(app, outputter) {
   if (!app.script)
     return new Error('No script path - aborting');
 
-  // Forbidden application name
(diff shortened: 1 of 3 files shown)


estliberitas:
@soyuka Em... I'm not sure, but without Automattic/cli-table#83 update tests won't work and whole commit may break PM2.

soyuka:
Indeed, but with no updates in 8 months and the amount of PR's we need to publish a fork.

estliberitas:
@soyuka :+1:

PR #1961, opened 2016-02-21
fix logging wrong "kill_timeout" into God#processIsDead

nskazki:
I just experimented and observed that the "real" time allotted for the process shutdown does not match recorded in the log.

in my app.json "kill_timeout" = 300 
```
{
  "name": "app",
  "script": ".",
  "kill_timeout": 300
}
```

but in the log I could see conflicting information:
(each `PM2 Process with pid 10696 still not killed, retrying...` = 100 ms)
```
PM2 Stopping app:app id:0
app-0 SIGINT
PM2 Process with pid 10696 still not killed, retrying...
PM2 Process with pid 10696 still not killed, retrying...
PM2 Process with pid 10696 still alive after 1600ms, sending it SIGKILL now...
```

so, pls accept my oneline fix :)

### DIFF
--- a/lib/God/Methods.js
+++ b/lib/God/Methods.js
@@ -228,7 +228,7 @@ module.exports = function(God) {
         return cb({type : 'timeout', msg : 'timeout'});
       }
       else {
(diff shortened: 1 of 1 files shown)


Unitech:
thanks :)

Issue #1750, opened 2015-11-05
pm2 install not working on windows


angelosc:
on windows run:
c:\> pm2 install pm2-logrotate

throws the following error:
[PM2][Module] Installing module pm2-logrotate
[PM2][Module] Processing...
The system cannot find the path specified.
[PM2][Module][ERROR] Unknown module

Tested on windows 7x64 and windows 2012R2x64

JamesBewley:
Same issue on windows 2012 R2

javahas:
I have the same issue, did you find a solution?

angelosc:
no, this is still not fixed

Mohinesh27:
I am also facing same issue. Please help me to out of it

jianbingfang:
+1

PR #2042, opened 2016-03-23
Replace shelljs "ugly hack" with shelljs support for silent console

BrianRosamilia:
Clean up shelljs npm install code for modules as shelljs supports a silent option

Additionally : pm2 modules are completely broken on Windows (and have been for quite some time) because of the /dev/null dependency.  This fixes https://github.com/Unitech/pm2/issues/1750

### DIFF
--- a/lib/Modularizer.js
+++ b/lib/Modularizer.js
@@ -105,9 +105,7 @@ function installModule(module_name, cb) {
       process.stdout.write('.');
     }, 500);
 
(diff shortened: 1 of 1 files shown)


BrianRosamilia:
If this isn't a workable fix for some reason, I plan to re-request with a require('os').platform === "win32" check to get pm2 modules working on windows.  Thanks.

PR #2073, opened 2016-04-06
Allows to update packages from tarball generated with 'npm pack'

flozz:
Hello,

I already made a PR some time ago (PR #1713) that allows to install module from .tgz packages (generated with the `npm pack` command). But there is a little issue: when updating a package with the `pm2 install foo.tgz` command, it does not updates the package and it starts a new instance of the already installed package. This PR fixes this issue.

Regards,


Unitech:
that works greatly, thanks

Issue #2071, opened 2016-04-05
pm2-dev @ 1.1.1 fails with Cannot find module '../lib/Log'


wulfsolter:
Node: v5.10.0
PM2: 1.1.1

Console:
    $ pm2-dev app.js

    module.js:341
        throw err;
        ^
    
    Error: Cannot find module '../lib/Log'
        at Function.Module._resolveFilename (module.js:339:15)
        at Function.Module._load (module.js:290:25)
        at Module.require (module.js:367:17)
        at require (internal/module.js:16:19)
        at Object.<anonymous> (/usr/local/lib/node_modules/pm2/bin/pm2-dev:10:17)
        at Module._compile (module.js:413:34)
        at Object.Module._extensions..js (module.js:422:10)
        at Module.load (module.js:357:32)
        at Function.Module._load (module.js:314:12)
        at Function.Module.runMain (module.js:447:10)

~/.pm2/pm2.log

    2014-05-26T23:39:53.925Z : Trace: [Error: todelete0 : id unknown]
        at Object.God.logAndGenerateError (/usr/local/share/npm/lib/node_modules/pm2/lib/God/Methods.js:26:15)
        at /usr/local/share/npm/lib/node_modules/pm2/lib/God/ActionMethods.js:211:30
        at Object.God.stopProcessId (/usr/local/share/npm/lib/node_modules/pm2/lib/God/ActionMethods.js:127:14)
        at Object.God.deleteProcessId (/usr/local/share/npm/lib/node_modules/pm2/lib/God/ActionMethods.js:210:9)
        at Worker.<anonymous> (/usr/local/share/npm/lib/node_modules/pm2/lib/God/Reload.js:121:13)
        at Worker.g (events.js:180:16)
        at Worker.EventEmitter.emit (events.js:95:17)
        at Object.messageHandler.listening (cluster.js:256:12)
        at handleMessage (cluster.js:197:32)
        at ChildProcess.EventEmitter.emit (events.js:117:20)
Works as intended @ 1.0.2

Unitech:
pm2-dev needs a total rewrite

brenc:
So...for those of us who are relatively new to NodeJS and PM2...I assume we're to use this as a workaround?

`pm2 start server.js --watch`

and:

`pm2 logs`

Which appears to work about the same as pm2-dev?

wulfsolter:
@brenc Yep, that's the easiest work around for now. 

[...]

PR #2075, opened 2016-04-08
fix(pm2-dev): fix Log module link

coogleyao:
fix Log module link


soyuka:
fix #2071 cc @Unitech LGTM for a temporary workaround

Unitech:
thanks, creating a a test file for that `pm2-dev` command would be nice too

Issue #2089, opened 2016-04-14
Programmatic API documentation please


fresheneesz:
[This](http://pm2.keymetrics.io/docs/usage/pm2-api/) isn't documentation. Those are examples. There is simply no substitute for exhaustive documentation that lists all options, parameters, and methods intended for public availability. I would offer my help, but I don't understand pm2's programatic because ... drum roll ... there's no docs. If someone that does understand the interface could please write some docs that list available functions, ALL the parameters those functions have, and their behavior, it would be much appreciated by your thousands of fans.

I see that under "Programmatic API" there is a list of methods, but they're woefully incomplete. They're also incredibly difficult to read. Some love there would be great.

soyuka:
Yes, we definitely need to put some time to improve this. The "programmatic api" is giving you access to every method of the [`CLI.js`](https://github.com/Unitech/pm2/blob/master/lib/CLI.js) file. It's complex, I won't deny that but @Unitech did a great work [separating concerns](https://github.com/Unitech/pm2/tree/master/lib/CLI) recently.
Every `options` parameters is defined in the [options schema](https://github.com/Unitech/pm2/blob/master/lib/CLI/schema.json), on which cli and json usage are based on.

To help further, you can take a look at a bunch of [examples](https://github.com/Unitech/pm2/tree/master/examples) but also [unit tests](https://github.com/Unitech/pm2/tree/master/test/programmatic) that rely on programmatic usage. I've also published an example that uses the programmatic api to notify by email when some events occurs. You can find it [here](https://github.com/soyuka/pm2-notify). Indeed, you'd have to dig a bit in the code...

As you must know there are a lot of options and possibilities with pm2. I've done some tutorials recently (https://github.com/pm2-hive/pm2-hive.github.io/pull/19, https://github.com/pm2-hive/pm2-hive.github.io/pull/15, https://github.com/pm2-hive/pm2-hive.github.io/pull/14), and will do my best to try  to improve this if no one else can help! 

Thanks for your interest in improving pm2 :).

fresheneesz:
I added some docs to the readme: https://github.com/Unitech/pm2/pull/2091

fresheneesz:
FYI I added a pull request here: https://github.com/pm2-hive/pm2-hive.github.io/pull/25

soyuka:
Closing this one in favor of https://github.com/pm2-hive/pm2-hive.github.io/pull/25

Note that it'll be merged soon, just waiting for @Unitech aproval.

PR #2091, opened 2016-04-14
Adding programmatic documentation

fresheneesz:
Because this is near-worthless: http://pm2.keymetrics.io/docs/usage/pm2-api/#programmatic-api and isn't accessible to github contributors. There is a lot that remains undocumented. What is the proc parameter in start, stop, and restart errbacks? What are all those options in the `start` command? What is graceful reload vs regular reload? What happens to console logs if you don't specify the `output` or `error` options? Where is the PID file if you don't specify that option? There are so many questions that are important and should be documented.

I completely ignored list, describe, dump, flush, reloadLogs, sendSignalToProcessName (omg that method name wtf), and startup - cause I don't know what any of them do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,58 @@ $ pm2 save
 
 [More about startup scripts](http://pm2.keymetrics.io/docs/usage/startup/)
 
+## Programmatic Usage
+
+You can also use pm2 directly from node.
+
+`npm install pm2 --save`
+
+**`pm2.connect(errback)`** - Connects to a running pm2 daemon ("God") or launches one.  
+
+* `errback(error)` - Called when finished connecting to or launching the pm2 daemon process.
+
+**`pm2.start(options, errback)`** - Connects to a running pm2 daemon ("God") or launches one.  
+**`pm2.start(jsonConfigFile, errback)`**  
+**`pm2.start(script, errback)`**  
+**`pm2.start(script, options, errback)`**  
+**`pm2.start(script, jsonConfigFile, errback)`**  
+
+* `script` - The path of the script to run
+* `options` - An object with the following options:
+  * `script` - The path of the script to run
+  * `exec_mode` - If set to 'cluster', will enable clustering (running multiple instances of the `script`)
+  * `instances` - (*Default: 1*) How many instances of `script` to create. Only relevant in `exec_mode` 'cluster'.
+  * `max_memory_restart` - If set and `script`'s memory usage goes about the configured number, pm2 restarts the `script`. Describe a number of megabytes with the suffix 'M', for example "100M". 
+  * `scriptArgs` - An array of arguments to start the script with.  
+  * `error` - The path to a file to append stderr output to.
+  * `output` - The path to a file to append stdout output to. 
+  * `pid` - The path to a file to write the pid of the started process. The file will be overwritten. Note that the file is not used in any way by pm2 and so the user is free to manipulate or remove that file at any time. The file will be deleted when the process is stopped or the daemon killed.
+  * `name` - An arbitrary name that can be used to interact with (e.g. restart) the process later.
+  * `watch` - If set to `true`, the application will be restarted on change of the `script` file.
+  * `nodeArgs`
+  * `cron`
+  * `mergeLogs`
+  * `runAsUser`
+  * `runAsGroup`
+  * `executeCommand`
+  * `interpreter`
+  * `write`
+* `jsonConfigFile` - The path to a JSON file that can contain the same options as the `options` parameter.
+* `errback(err,proc)` - An errback called when the `script` has been started. 
+
+**`pm2.disconnect()`** - Disconnects from the pm2 daemon.
+
+**`pm2.stop(process, errback)`** - Stops a process.  
+**`pm2.restart(process, errback)`** - Restarts a process.  
+**`pm2.delete(process, errback)`** - ?  
+**`pm2.reload(process, errback)`** - ?  
+**`pm2.gracefulReload(process, errback)`** - ?  
+
+* `process` - Can either be the `name` as given in the `pm2.start` `options`, a process id, or the string "all" to indicate that all scripts should be restarted.
+* `errback(err, proc)`
+
+**`pm2.killDaemon(errback)`** - Kills the pm2 daemon. Note that when the daemon is killed, all its processes are also killed. Also note that you still have to explicitly disconnect from the daemon even after you kill it.
+
 ## Keymetrics monitoring
 
 [![Keymetrics Dashboard](https://keymetrics.io/assets/images/application-demo.png)](https://app.keymetrics.io/#/register)


soyuka:
> Because this is near-worthless: http://pm2.keymetrics.io/docs/usage/pm2-api/#programmatic-api and isn't accessible to github contributors.

The repository for documentation is available at https://github.com/pm2-hive/pm2-hive.github.io/ ;).

> What is the proc parameter in start, stop, and restart errbacks? 

This is a full process object.

> What are all those options in the start command?

Already answered in #2089, take a look at the schema

> What is graceful reload vs regular reload?

This is explained [here](http://pm2.keymetrics.io/docs/usage/pm2-doc-single-page/#graceful-reload)

> What happens to console logs if you don't specify the output or error options?

They are redirected to default log files, written by default in `~/.pm2/logs/app_name.err|out.log`

> Where is the PID file if you don't specify that option?

Defaults to `~/.pm2/pids/app_name-id.pid` 

> `sendSignalToProcessName`

I don't see what's wrong, it's self-explanatory.

fresheneesz:
@soyuka Ah good to know there is a place I can contribute to the docs on the website. Shall I move this documentation there? 

Thanks for all the answers, I'll document them somewhere if you'll accept a pull request. 

`sendSignalToMethodName` - Its just a long name lol. But in any case, its not self explanatory to me. Of course you can't send a signal to a "name" and I don't understand what it would mean to send to a "method". 

As for graceful reload, that helps, but it still leaves me wondering how exactly `restart` and `stop` stop the process. I assume that your normal `sigterm` signal is used to kill the process for those, which also allows graceful program termination, but the existence of an alternative graceful reload mechanism leaves me worried that `stop` and `restart` kill a program instantly without letting it properly shut down. Because of this, I think its especially important to document how `stop` and `restart` work in this respect.

soyuka:
> Shall I move this documentation there?

This would be better indeed! Thanks.

> I don't understand what it would mean to send to a "method".

Here it's `sendSignalToProcessName`, no method. This is a high level method that allows you to send a signal to a pm2-named process instead of the usual process id.

> I think its especially important to document how stop and restart work in this respect.

Take a look at [signals and clean restarts](http://pm2.keymetrics.io/docs/usage/signals-clean-restart/#signals). pm2 will send a SIGKILL after some time when using `stop`/`restart`.

fresheneesz:
Ok, I'm in the process of rewriting the docs page at pm2-hive.github.io . I have more questions though:

> This is a full process object.

You mean a [child process](https://nodejs.org/api/child_process.html) object?

> Every options parameters is defined in the [options schema](https://github.com/Unitech/pm2/blob/master/lib/CLI/schema.json)

I see `node_args` but not `nodeArgs` in there, are they both equally valid? FYI, its far less common in javscript to use underscores - most people use camelCase. I'd recommend sticking to camelCase. I also see `args` defined in that schema, but its undocumented. What's the difference between `args`, `nodeArgs`, and `scriptArgs`?

I also see things like `log_file`, `error_file`, `out_file`, and `pid_file`, but not `error`, `output`, and `pid`. Why is there inconsistency between the docs and this schema file? Same question about `interpreter` vs `exec_interpreter`. 

Where is documentation on how to write a `log_date_format`? 

What does it mean that `merge_logs` will "merge logs from all instances of the same app into the same file"? How does it determine what the "same app" is? 

I still don't understand `reload` - but I linked to the section where it's explained. It would be great if that section could be expanded to explain how exactly this works (possibly with references to the [node cluster documentation](https://nodejs.org/docs/latest/api/cluster.html)).

I experimented with the pid file, and found that "the file is not used in any way by pm2 and so the user is free to manipulate or remove that file at any time." Am I right here?

What is "no daemon" mode? What does it have to do with passing `true` as the first argument to `connect`? 

What is the `port` option? What is the `log_file` option? 

What does `pm2.delete`/`pm2.dump`/`pm2.reloadLogs` do? What's the output format for `pm2.list` and `pm2.describe`? 

Is the `bus` returned from `launchBus` an [EventEmitter](https://nodejs.org/api/events.html) object or something else? 

What does `pm2.startup` do? What is its "platform" parameter?

It seems like pm2 has lots of unnecessary or redundant options. For example, if a ChildProcess object is passed back from start, the pid can be accessed directly from there and a file written if the user wants. And there is no reason to allow a user to pass the path to a json config file OR a javascript object - just let them load a json config file themselves and pass it in as a javascript object.  And why is the `next_gen_js` option needed if you can just pass in the `--harmony` node argument? Also, something like the watch feature is easily doable with other modules, its really not something that pm2 should be handling itself.

soyuka:
Nope, the `proc` object in pm2 looks [like this](https://github.com/soyuka/pm2-notify#templating).

> I see node_args but not nodeArgs in there, are they both equally valid?

Those are computed the same way when the configuration comes from a json file. Having both originally comes from having CLI parameters that are usually not camel-cased.  

 > What's the difference between args, nodeArgs, and scriptArgs?

[From a stackoverflow answer of mine](http://stackoverflow.com/questions/29900253/debug-application-which-is-run-using-pm2/29900353#29900353):

> - args are your script arguments
- node_args are arguments that are passed to node executable

And I think those are documented [here](http://pm2.keymetrics.io/docs/usage/application-declaration/#attributes-available)

> I also see things like log_file, error_file, out_file, and pid_file, but not error, output, and pid. Why is there inconsistency between the docs and this schema file? 

We stick with `log_file`, `out_file`, `pid_file` when speaking about a json schema declaration file. But, when it comes to command line usage, it's easier to just tell pm2 `--e error.log`.

> Same question about interpreter vs exec_interpreter.

Those are just aliases. They were added very recently, along with `interpreter_args` that allows you to pass arguments to `python` the same way you'd use `node_args`.

> Where is documentation on how to write a log_date_format?

It's a moment format: http://momentjs.com/.

> What does it mean that merge_logs will "merge logs from all instances of the same app into the same file"? How does it determine what the "same app" is?

They are always misunderstandings with `merge_logs`, what it simply does is that it combines your processes logs into one. For example say you launch a cluster of 4 processes (`pm2 start -i 4 test.js`), you'll have 4 different logs (actually 8, 4 out logs, 4 err logs). When using `--merge-logs` pm2 will combine those in one file (here 2, 1 out log, 1 err log).

> I still don't understand reload

It does a restart by keeping at least one process alive so that the server stays alive. There are no real link to the nodejs cluster documentation because it's an intern logic. Yes, it requires the cluster module to work but that's all.

> it seems like since pm2 is a daemon, it could store all the pids in memory and not need to use a pid file.

Yes it does and don't require pid files. It still writes pid files in case you need one from an outside position. IMO it's handy to `cat ~/.pm2/pids/my-app.pid` an get your pid.

> What is "no daemon" mode? 

`no-daemon` mode is a "stay-alive" process. Launch it in CLI and you'll see that pm2 runs in foreground, not daemonized. A SIGINT signal will kill pm2 along with your running script.

> What does it have to do with passing true as the first argument to connect?

Because `CLI.connect` is [`Satan.start`](https://github.com/Unitech/pm2/blob/79ba743723131436bd049cc952e2b1f1158bc84c/lib/Satan.js#L40). Satan handles connections to the Daemon through RPC (God). It also launches the daemon, either in a sub-process, either in the same process as the current one. 

> What is the port option? 

Port option is useless, it was added because some wanted to see their application port in the `pm2 list`. It has been removed since, maybe not removed enough.

> What is the log_file option?

`log_file` is to specify one log file for both `stdout` and `stderr` combined. Now, if you had `merge_logs` and `log_file` on the previous cluster server (`pm2 start -i 4 --merge-logs --log-file test.log server.js`) you'll have only one log containing every processes with stdout and stderr mixed.

> What does pm2.delete/pm2.dump/pm2.reloadLogs do? What's the output format for pm2.list and pm2.describe?

```
delete <name|id|script|all|json|stdin...>                            stop and delete a process from pm2 process list
dump                                                                 dump all processes for resurrecting them later
describe <id>                                                        describe all parameters of a process id
reloadLogs                                                           reload all logs
```

Reload logs just rotate everything.

If you provide a callback to either of those methods (`list` or `describe`), you'll get an `array` of `objects` with `list` and a single process `object` with `describe`. `list` should be an array of `describe` results.

> Is the bus returned from launchBus an EventEmitter object or something else?

It's an [axon sub emitter](https://github.com/tj/axon#pubemitter--subemitter) ([source code ref](https://github.com/Unitech/pm2/blob/79ba743723131436bd049cc952e2b1f1158bc84c/lib/Satan.js#L452)).

> What does pm2.startup do? What is its "platform" parameter?

This tries to create a startup file according to your platform:

```
❯ pm2 startup --help

  Usage: startup [options] [platform]

  auto resurrect process at startup. [platform] = ubuntu, centos, redhat, gentoo, systemd, darwin, amazon

  Options:

    -h, --help  output usage information
```

>  And why is the next_gen_js option needed if you can just pass in the --harmony node argument? 

This option has been deprecated recently, in favour of [different ways to handle babel](http://pm2.keymetrics.io/docs/tutorials/using-transpilers-with-pm2). And I agree with you about the harmony flag, but not anyone can stick with basic ES6 features.

> Also, something like the watch feature is easily doable with other modules, its really not something that pm2 should be handling itself.

Yes but it has been requested as a feature by many users. You don't have to use it, but it's embed if you want it.

> It seems like pm2 has lots of unnecessary or redundant options. (...) And there is no reason to allow a user to pass the path to a json config file OR a javascript object - just let them load a json config file themselves and pass it in as a javascript object.

You have to know that there's a lot of history too. At first, it was only CLI, then json declaration files (which can be sent to pm2's stdin on top of a file path). Then only came the programmatic api which is only a substract of CLI.js. There has been work to refactor those kind of things, but it's hard not to break anything when you support as many features as pm2 does.

Hope this helps!

fresheneesz:
@soyuka Thanks! This is all very helpful! I have more questions still tho:

> * args are your script arguments
> * node_args are arguments that are passed to node executable

What about `scriptArgs`?

Can anything that is written with underscores (like `log_file`) be written in camel case?

> We stick with log_file, out_file, pid_file when speaking about a json schema declaration file. But, when it comes to command line usage, it's easier to just tell pm2 --e error.log.

So `error` and `err_file` are equally valid in the json config object as well?

> mergeLogs

If you explicitly specify a log file via `error` or `output`, how does pm2 create log files? Does it merge them even if you don't set `mergeLogs` to true or does it do something like append a number on the end of your filename?

> pm2.delete/pm2.dump

I'm still not sure I understand what these do. So if you only `stop` a process, it remains in pm2's list? But deleting removes it as if it were never there? What is the result of dump and how is that result used? 

> pm2.reloadLogs

So this creates new log files and starts appending to the new files instead of the old ones? How does this relate to the names you may have defined in `output`, `error`, or `log_file`? What happens if you kill pm2 and restart all the processes with their original configuration? Will this whole rotation process be reset so that the original log files will start filling up again?

> pm2.startup

Oh that is very interesting. How does this work? What script does it startup on machine boot? What if the machine crashes  unexpectedly, will pm2 still startup correctly?

> describe

I went through the output to describe, and I was wondering what `pm_id` is. Also, what does the value of `monit.cpu` mean? Is that its "nice" value or something? In mine it says '7'. Also, what are the possible `status` values?

> Yes but it has been requested as a feature by many users.

This isn't a good reason to implement something. Those users should be directed to something like [watchpack](https://github.com/webpack/watchpack) or [chokidar](https://github.com/paulmillr/chokidar) or whatever module pm2 is using to do this. I want to suggest that that functionality be deprecated with some recommendations put in the docs for how to do it with other modules.

> You have to know that there's a lot of history too.

Gotcha.

fresheneesz:
I made a pull request for a new API page: https://github.com/pm2-hive/pm2-hive.github.io/pull/25 , so i'll close this one.

soyuka:
> What about scriptArgs?

Those are your script arguments. If they are not specified, pm2 takes every argument after `--`.

> Can anything that is written with underscores (like log_file) be written in camel case?

Yes, but it's better to document a camelCase usage for options relative to the programmatic usage.

> So error and err_file are equally valid in the json config object as well?

Indeed.

> If you explicitly specify a log file via error or output, how does pm2 create log files? Does it merge them even if you don't set mergeLogs to true or does it do something like append a number on the end of your filename?

By default it's creating a log of the format `${process.name}-${out|err}-${number}.log` for each output fd: `stderr`  `stdout`. 

> I'm still not sure I understand what these do. So if you only stop a process, it remains in pm2's list? But deleting removes it as if it were never there? 

Yes, `delete` removes it from pm2 memory.

> What is the result of dump and how is that result used?

`dump` writes the process list in a json format to the file specified in `DUMP_FILE_PATH` (defaults to     ~/.pm2/dump.pm2) [declared in the constants file](https://github.com/Unitech/pm2/blob/master/lib/samples/sample-conf.js).  

> So `pm2.reloadLogs` creates new log files and starts appending to the new files instead of the old ones? How does this relate to the names you may have defined in output, error, or log_file? 

Yes. It just adds a number at the end of the file and starts back to log in an empty file. Name doesn't change for current log. 

> What happens if you kill pm2 and restart all the processes with their original configuration? Will this whole rotation process be reset so that the original log files will start filling up again?

Old logs will stay, but default files will start to fill up again yes.

> `pm2.startup` Oh that is very interesting. How does this work? What script does it startup on machine boot? What if the machine crashes unexpectedly, will pm2 still startup correctly?

Simply by adding the correct [startup script](https://github.com/Unitech/pm2/tree/79ba743723131436bd049cc952e2b1f1158bc84c/lib/scripts) to the system. For example with the [`update-rc.d` script](https://github.com/Unitech/pm2/blob/79ba743723131436bd049cc952e2b1f1158bc84c/lib/scripts/pm2-init.sh#L60), it'll launch `pm2 resurrect` (resurrect previously dumped processes)

> I went through the output to describe, and I was wondering what pm_id is. Also, what does the value of monit.cpu mean? Is that its "nice" value or something? In mine it says '7'. Also, what are the possible status values?

`pm_id` is pm2's internal id for a process. It's kept in memory and can be used as parameter for `restart, delete, describe, logs etc.`. The `monit` values are cpu % usage and memory usage in bytes, see also [pidusage](https://github.com/soyuka/pidusage).

For possible status values ([listed here](https://github.com/Unitech/pm2/blob/02df361d39419224dff1563f38fb2c3dd73e75c3/constants.js#L58)):

```
  ONLINE_STATUS          : 'online',
  STOPPED_STATUS         : 'stopped',
  STOPPING_STATUS        : 'stopping',
  LAUNCHING_STATUS       : 'launching',
  ERRORED_STATUS         : 'errored',
  ONE_LAUNCH_STATUS      : 'one-launch-status',
```

> This isn't a good reason to implement something. Those users should be directed to something like watchpack or chokidar or whatever module pm2 is using to do this. I want to suggest that that functionality be deprecated with some recommendations put in the docs for how to do it with other modules.

We're using chokidar. There's almost nothing on pm2's side, only high-level api's to allow the user to watch paths related to specific processes, or applications. Pm2 is also useful for development, for example `pm2-dev` that watches, streams logs on the fly.
I personally agrees with you on the subject, but I did the watch implementation myself because in some cases, it's really useful. We're not going to deprecate it ;).

fresheneesz:
> Those are your script arguments. If they are not specified, pm2 takes every argument after --.

I don't quite understand here. Are `scriptArgs` the same as `args` then? What do you mean about `--` - what's an example there?

I added everything else into a new revision of my push request. Thanks!

soyuka:
For example in `pm2 start app.js -- --someoption -f someargument`, script arguments would be ` --someoption -f someargument`.

Added a small comment there and we're good to go. Thanks to you!

fresheneesz:
@soyuka No problem! And thanks for your informative help!

PR #2096, opened 2016-04-15
fix cli for node v6

targos:
When `conf.script` is `undefined`, `p.basename` currently returns `"undefined"` in Node.js v5.x. v6.0.0 throws an exception in this case. I'm not sure what it is expected to be so I reproduced the behavior.
Perhaps it should be the empty string instead ?

### DIFF
--- a/lib/CLI.js
+++ b/lib/CLI.js
@@ -2369,7 +2369,7 @@ function checkDeprecates(conf){
  */
 function prepareAppName(conf){
   if (!conf.name){
-    conf.name = p.basename(conf.script);
+    conf.name = conf.script !== undefined ? p.basename(conf.script) : 'undefined';
     var lastDot = conf.name.lastIndexOf('.');
     if (lastDot > 0){
       conf.name = conf.name.slice(0, lastDot);


lib/CLI.js
```diff
@@ -2369,7 +2369,7 @@ function checkDeprecates(conf){
  */
 function prepareAppName(conf){
   if (!conf.name){
-    conf.name = p.basename(conf.script);
+    conf.name = conf.script !== undefined ? p.basename(conf.script) : 'undefined';
```
soyuka: can't this be fixed on a higher level? Also, having `conf.name = 'undefined'` can't be correct.

targos: Yes, I agree. I'm not familiar with your code base though so I'm open to any suggestion on how to fix it properly.

soyuka: How did you trigger this error? `script` should always be defined IMO, maybe we can throw a ReferenceError if it isn't.

targos: Just executing `pm2 kill`:

From the repo (no running process):
```
pm2 [master●] % ./bin/pm2 kill
[PM2] Spawning PM2 daemon
[PM2] PM2 Successfully daemonized
[PM2] Stopping PM2...
path.js:7
    throw new TypeError('Path must be a string. Received ' + inspect(path));
    ^

TypeError: Path must be a string. Received undefined
    at assertPath (path.js:7:11)
    at Object.basename (path.js:1355:5)
    at prepareAppName (/home/mzasso/git/forks/pm2/lib/CLI.js:2373:19)
    at verifyConfs (/home/mzasso/git/forks/pm2/lib/CLI.js:2296:5)
    at Object.CLI._handleAttributeUpdate (/home/mzasso/git/forks/pm2/lib/CLI.js:1187:17)
    at Object.CLI._operate (/home/mzasso/git/forks/pm2/lib/CLI.js:1016:16)
    at /home/mzasso/git/forks/pm2/lib/CLI.js:1677:11
    at /home/mzasso/git/forks/pm2/lib/CLI.js:1749:19
    at /home/mzasso/git/forks/pm2/node_modules/async/lib/async.js:52:16
    at replenish (/home/mzasso/git/forks/pm2/node_modules/async/lib/async.js:314:29)
```

soyuka: I need to investigate this, thanks for reporting.

Unitech:
bug reproduced with latest nodejs 6 build, this fix the issue, thanks

PR #2116, opened 2016-04-28
Update VersionManagement.js

jonaswindey:
Add error code when command fails

### DIFF
--- a/lib/tools/VersionManagement.js
+++ b/lib/tools/VersionManagement.js
@@ -291,7 +291,7 @@ var execCommands = function(repo_path, command_list, cb) {
       if (code === 0)
         callback();
       else
-        callback('`'+command+'` failed');
+        callback('`'+command+'` failed with error code ' + code);
     });
   }, function(err) {
     if (err)


lib/tools/VersionManagement.js
```diff
@@ -291,7 +291,7 @@ var execCommands = function(repo_path, command_list, cb) {
       if (code === 0)
         callback();
       else
-        callback('`'+command+'` failed');
+        callback('`'+command+'` failed with error code ' + code);
```
soyuka: I would also add command stderr to the error message, WDYT?

Unitech:
@jonaswindey would you mind to do this PR again on PM2 v2: https://github.com/Unitech/pm2/blob/79ba743723131436bd049cc952e2b1f1158bc84c/lib/tools/VersionManagement.js#L294

thanks

Unitech:
@jonaswindey would you mind to do this PR again on PM2 v2: https://github.com/Unitech/pm2/blob/79ba743723131436bd049cc952e2b1f1158bc84c/lib/tools/VersionManagement.js#L294

thanks

Issue #1562, opened 2015-08-28
Daemon should support non-background mode


wmertens:
For process monitoring services like launchd, systemd and monit, it would be best if you could run the PM2 daemon without it going to the background.

Then all the OS tools can be used to set resource limits, do monitoring etc.

soyuka:
`--no-daemon` would suffice?

wmertens:
Oh, I didn't find that. In that case, shouldn't the startup scripts for
systemd and launchd use that option?

On Fri, Aug 28, 2015 at 2:00 PM Antoine Bluchet <notifications@github.com>
wrote:

> --no-daemon would suffice?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Unitech/PM2/issues/1562#issuecomment-135749703>.
>
-- 

Wout.
(typed on mobile, excuse terseness)

soyuka:
I won't see why they should but I'm all ears if you have pros arguments.

wmertens:
Well the main argument is that launchd will know that PM2 is running and it
will show like that when running `launchctl list`.
Systemd can work around the forking but non-forking is preferred and
recommended.

On Fri, Aug 28, 2015 at 7:06 PM Antoine Bluchet <notifications@github.com>
wrote:

> I won't see why they should but I'm all ears if you have pros arguments.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Unitech/PM2/issues/1562#issuecomment-135833699>.
>
-- 

Wout.
(typed on mobile, excuse terseness)

PR #2129, opened 2016-05-03
Load modules in no-daemon mode

fauxsoup:
I don't know if this is intended behavior or not, but no-daemon mode prevents modules from being loaded.

Some background because I never filed an issue: I'm using PM2 with Windows services, which requires no-daemon mode (otherwise Windows thinks the service has crashed when the CLI process exits). I also want to use pm2-logrotate to keep the disk usage down. I imagine this would also be useful with respect to the discussion in #1562.


Unitech:
merged in development branch

Issue #2133, opened 2016-05-06
pm2 goes to infinite restart on child.process.exec('pm2 reload other_task')


fend25:
Hi. Thanks for pm2.

So, there are two tasks started under pm2: task1 and task2.

pm2 goes to infinite restart loop (on task1) when the task1 performs `child_process.exec('pm2 reload task2')`. Task2, by the way, successfully reloads (or restarts, with 'restart' command there is the same result). Although it switches both tasks into watching mode. `--no-autorestart` and `--no-automation` keys don't help.

I created a repo with minimal sample how to reproduce error: https://github.com/fend25/test_pm2_reload

Technical details::
os: ubuntu~14.04.1 x86_64 3.19.0-47-generic
node: v6.0.0
pm2: v1.1.3

Tasks have been started in fork mode, No clusters or so on.

Cyral:
+1 I think I am having the same issue, using a pm2 process to restart other processes by running a bash script, process 1 is stopped successfully running "pm2 restart 1" but then the process that executed the script is ran again.

Example: 

spawn("sh", ["update.sh"]);
...
**update.sh:**
echo Updating application
git pull
pm2 restart 1

**Output:**
```
PM2 [Watch] Start watching 1
PM2 Stopping app:<restarted app> 1 id:1
PM2 App [<restarted app>] with id [2] and pid [3425], exited with code [2] via signal [SIGINT]
PM2 Starting execution sequence in -fork mode- for app name:<original app> id:5
PM2 App name:<original app> id:5 online
```

<restarted app> is the app I am trying to restart, but <original app> is started instead of the app that should be restarted.

Cyral:
Looks like it might be related to #783

Cyral:
Okay so a few things... 

The old fix (that has since been removed) for this issue ([here](https://github.com/Unitech/pm2/pull/782/files)) is incorrect I believe because it restarts the app with a new ID, not the old ID (which breaks restart scripts and such).

Note that the issue also occurs if using the pm2 programmatic API.

I'm really not familiar enough with the pm2 internals to continue trying to fix this, but the issue is that the pm2_id in the environment variable is incorrect because it is using the parents environment variables.

Unitech:
Some environment variables are messing with the current process,
A temporary fix:

```javascript
  child_process.exec(`pm2 reload test_pm2_main`, { env : {
    PATH : process.env.PATH,
    HOME : process.env.HOME
  }}, function(err, stderr, stdout) {
[...]

Issue #783, opened 2014-10-27
Unable to restart pm2 process from within another pm2 process (0.11.1)


edorivai:
For quite a while I've been running a production server using PM2 0.9.1. We run two processes; let's call them *restartable.js* and *restarter.js*.

*restartable.js* is a script that will be restarted by *restarter.js*. This is done by running a child process withing *restarter.js* like this: 
```javascript
var spawnChildProcess = require('child_process').spawn;
spawnChildProcess('pm2', ['restart', '0']); // 0 is the identifier of restartable.js
```
This all worked flawlessly in 0.9.1, but since we update to 0.11.1 we're unable to restart *restartable.js* through *restarter.js*. What happens is that *restartable.js* is stopped, but never actually restarted.
The following is printed when logging for *restartable.js* right after the restart attempt
```
[PM2] Process with pid 41519 still not killed, retrying...
[PM2] Process with pid 41519 killed
[PM2] Starting execution sequence in -fork mode- for app name:restarter id:1
[PM2] App name:restarter id:1 online
```

From that point on it is also not possible to restart the process through commandline with
```
pm2 restart 0
```
I've also tried the same setup using pm2@0.10.0; this showed the same problem.
We've tried this setup with the most basic content for *restartable.js*, which ruled out any problems specific to the restartable script.

Is this the expected behaviour? We've reverted to 0.9.1 for now, but I'm kinda suspecting I'm misusing PM2.

Unitech:
Fixed with the lastest PM2 (>0.12.x)

PR #2137, opened 2016-05-06
Fix incorrect process from starting after a PM2 controlled process re…

Cyral:
…starts another PM2 process.

See #2133, appears to be a regression of #783


Cyral:
Looks like the fix done before is incorrect as it creates a new ID when it should use the old ID... going to have to look into this a bit deeper.

PR #2194, opened 2016-05-29
livescript support

### LINKED ISSUE #701: Add LiveScript support
Support for LiveScript (http://livescript.net) would be awesome.

dex4er:
This patch adds support for LiveScript (http://livescript.net) so it is possible to start LiveScript script in cluster mode.

It also should close #701


soyuka:
👍

dex4er:
If you don't like dependency on livescript then just put it as dev-dependency. I think it might also work for coffeescript.

coffee and live scripts should be provided with those packages already so pm2 might not require them.

dex4er:
Repatched to development branch

Unitech:
Nice, is that possible to install a global npm package that includes dev dependencies?

dex4er:
@Unitech I moved livescript dependency to devDependencies, because usually you install livescript (or coffescript) explicitly (and with `-g` option) and `require("livescript")` in pm2 is used lazily. 

Then I think livescript package is required only for tests of pm2.

In other words: you don't install Python and Perl by default even if pm2 does support them.

soyuka:
I agree, but then we should also move the `coffee` dependency to `dev`. Btw, if we can't require them, we  need to catch the error and display a meaningful message.

dex4er:
Resubmitted as pull #2248

PR #2420, opened 2016-09-30
Do not override app config with opts default values (fix #2366)

### LINKED ISSUE #2366: pm2-dev watch json options ignored on 2.0.7 - win7
I have following json conf file:

````
{
  "apps": [
    {
      "name": "dev-web-app",
      "script": "index.js",
      "watch": [
        "server",
        "shared",
        "srv_lib"
      ],
      "merge_logs": true,
      "env": {
        "NODE_ENV": "development",
        "NODE_PATH": ""
      },
      "exec_mode": "fork_mode"
    }
}
````

Then I run following command

````
pm2-dev start config/pm2/_development.json
````

#### Expected behaviour

I expect pm2 restarts process when files changed in following dir:

````        
"server",
"shared",
"srv_lib"
````

#### Actual behaviour

Process is restarted when a file is changed in all project folder

#### Steps to reproduce
Problem is here: https://github.com/Unitech/pm2/blob/development/lib/API.js#L851-L861

command options override app options insecurely

I can fix this issue if I make this change

````
  appConf.forEach(function(app) {
    if (opts.only && opts.only != app.name)
      return false;
    if (opts.watch && opts.watch === true && !app.watch)
      app.watch = true;
    if (opts.ignore_watch && !app.ignore_watch) // same issue with ignore_watch
      app.ignore_watch = opts.ignore_watch;
    if (opts.instances && typeof(opts.instances) === 'number')
      app.instances = opts.instances;
    apps_name.push(app.name);
  });
````

#### Software versions used

```
OS         : win7
node.js    : 6.5.0
PM2        : 2.0.7
```

### PR DESCRIPTION
**Pull request by danielbayerlein**
Fix pm2-dev issue - #2366 
Do not override the application config with the opts default values.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2366
| License       | MIT

### DIFF
--- a/lib/API.js
+++ b/lib/API.js
@@ -851,9 +851,9 @@ API.prototype._startJson = function(file, opts, action, pipe, cb) {
   appConf.forEach(function(app) {
     if (opts.only && opts.only != app.name)
       return false;
-    if (opts.watch && opts.watch === true)
+    if (opts.watch && opts.watch === true && !app.watch)
       app.watch = true;
-    if (opts.ignore_watch)
+    if (opts.ignore_watch && !app.ignore_watch)
       app.ignore_watch = opts.ignore_watch;
     if (opts.instances && typeof(opts.instances) === 'number')
       app.instances = opts.instances;
--- a/lib/API.js
+++ b/lib/API.js
@@ -851,9 +851,9 @@ API.prototype._startJson = function(file, opts, action, pipe, cb) {
   appConf.forEach(function(app) {
     if (opts.only && opts.only != app.name)
       return false;
-    if (opts.watch && opts.watch === true && !app.watch)
+    if (!app.watch && opts.watch && opts.watch === true)
       app.watch = true;
-    if (opts.ignore_watch && !app.ignore_watch)
+    if (!app.ignore_watch && opts.ignore_watch)
       app.ignore_watch = opts.ignore_watch;
     if (opts.instances && typeof(opts.instances) === 'number')
       app.instances = opts.instances;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: lib/API.js [hunk 1]
```diff
@@ -851,9 +851,9 @@ API.prototype._startJson = function(file, opts, action, pipe, cb) {
   appConf.forEach(function(app) {
     if (opts.only && opts.only != app.name)
       return false;
-    if (opts.watch && opts.watch === true)
+    if (opts.watch && opts.watch === true && !app.watch)
       app.watch = true;
-    if (opts.ignore_watch)
+    if (opts.ignore_watch && !app.ignore_watch)
```
soyuka: can you reverse the check and test `!app.watch` before `opts` tests? Will be faster in that case ;). Same for the second condition `(!app.ignore_watch && opts.ignore_watch)`.

Maybe we can add a test [here](https://github.com/Unitech/pm2/blob/master/test/bash/pm2-dev.sh) to prevent regression?

danielbayerlein: @soyuka Thank you for your feedback! Have a look at 53c814d and 86af1f5. I'm not sure why the Node.js 0.10 was not successful. I think it's a timing issue. Can you restart the Node.js 0.10 test?

vmarchaud: @danielbayerlein Restarted, wait & see ;)

danielbayerlein: @vmarchaud Thank you 😉

danielbayerlein:
@Unitech Ready to merge?

soyuka:
ping @vmarchaud

danielbayerlein:
@soyuka Thanks a lot!

Issue #2353, opened 2016-08-22
QUESTION: Tell pm2 when server is actually ready to handle requests


4storia:
I've looked around the documentation, but I didn't really see any references to this, so my apologies if this is covered somewhere already.

I have a koa app that mounts many sub-apps before coming online. Because it has to require many (around 15) sub-apps, it takes 5-10 seconds to start/restart fully, aka to get to the point where it is actually ready to accept requests.  It doesn't seem like pm2 picks up on this fact (not that I would expect it to), and when I do `pm2 reload`, there is a period of time where requests fail, as the application has not yet started listening for requests. Is there a way for me to tell pm2, from my application, that a process is *actually* "online"? Or is there a suggested workflow for reloading the application (which I'm running in cluster mode) so that my users don't experience downtime during deployments?

Unitech:
You may want to call the .listen function once all your sub applications are ready. PM2 recognize an application as ready once the app calls .listen.

You may also want to override the kill timeout:
http://pm2.keymetrics.io/docs/usage/signals-clean-restart/#customize-exit-delay

Let me know if it helps,

4storia:
So i'm already doing the following:
```
var app = koa();
... (requiring many sub-apps)
console.log("calling listen")
server = app.listen(port);
....
module.exports = app;
```

While tailing all pm2 logs, I see:
```
2016-08-23-13:32:26 PM2 Starting execution sequence in -fork mode- for app name:app id:0
2016-08-23-13:32:26 PM2 App name:app id:0 online
...
(several console.log statements about mounting sub-apps)
...
2016-08-23-13:32:32 app-0 calling listen
2016-08-23-13:32:32 app-0 Server is running.  { address: '::', family: 'IPv6', port: 8080 }
```

As you can see, pm2 says my application is "online" long before `app.listen` is actually executed. The above output is from operating without cluster mode for the sake of debugging, but the outcome is the same with cluster mode enabled. 
I'm not particularly interested with the exit of the application (at least not at the moment), because my problem is really that pm2 lists my application as "online" as soon as `app.js` begins execution, rather than when `app.listen` is called.

I couldn't seem to find where pm2 specifically listens for `app.listen`, could you point me to the general area of code where this happens?

vmarchaud:
[...]

PR #2472, opened 2016-10-26
Wait ready event to set application as started

vmarchaud:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2353
| License       | MIT

when passing option `--wait-ready` in CLI or `wait_ready` in process file, pm2 will wait for the application to `process.send('ready');` to set it as `started`


lib/God/ForkMode.js
```diff
@@ -105,7 +105,6 @@ module.exports = function ForkMode(God) {
       cspr.process = {};
       cspr.process.pid = cspr.pid;
       cspr.pm2_env = pm2_env;
-      cspr.pm2_env.status = cst.ONLINE_STATUS;
```
Unitech: why?

vmarchaud: When the application is started, the status is assigned only after the application is online, not before.

soyuka: Can't we keep this but put a `cst.LAUNCHING_STATUS` or something like this?

Unitech:
it works well, we should add some sort of timeout to this feature. Like passing an optional timeout time to the `--wait-ready` option

PR #2526, opened 2016-11-16
Added launch modules function to api.

bwkiefer:
Hi, I added this function to start up modules programatically, because using API.connect with no daemon set to true eventually results in the modules not being launched at [line 134 of API.js](https://github.com/Unitech/pm2/blob/development/lib/API.js#L136)
```javascript
if (meta.new_pm2_instance == false)
  return cb(err, meta);
```

Cheers!

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/lib/API.js
+++ b/lib/API.js
@@ -203,6 +203,15 @@ API.prototype.disconnect = API.prototype.close = function(cb) {
 };
 
 /**
+ * Launch modules
+ *
+ * @param {Function} cb callback once pm2 has launched modules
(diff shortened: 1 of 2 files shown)


Unitech:
Great! Thanks

PR #2569, opened 2016-12-12
Improve opts.watch

pi0:
This PR Simply checks if watch option is string form of "true", this typo may happen when developers are using `echosystem.json` file and may something like this: `"watch":"true"`  

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

### DIFF
--- a/lib/API.js
+++ b/lib/API.js
@@ -870,8 +870,8 @@ API.prototype._startJson = function(file, opts, action, pipe, cb) {
   appConf.forEach(function(app) {
     if (opts.only && opts.only != app.name)
       return false;
-    if (!app.watch && opts.watch && (opts.watch === true || opts.watch === "true") )
-      app.watch = true;
+    if (!app.watch && opts.watch)
+      app.watch = JSON.parse(opts.watch);
     if (!app.ignore_watch && opts.ignore_watch)
       app.ignore_watch = opts.ignore_watch;
     if (opts.instances && typeof(opts.instances) === 'number')
--- a/lib/API.js
+++ b/lib/API.js
@@ -870,7 +870,7 @@ API.prototype._startJson = function(file, opts, action, pipe, cb) {
   appConf.forEach(function(app) {
     if (opts.only && opts.only != app.name)
       return false;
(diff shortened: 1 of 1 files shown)


soyuka:
I do not agree with this as 

```
{
    "watch": true
}
```

Is valid json you should use a boolean, not a string.

pi0:
@soyuka Thats true, but consider also 
```js
{
      "watch": "true"
}
```
Is a valid json! However The only as mentioned in [PM2 docs](http://pm2.keymetrics.io/docs/usage/application-declaration/#advanced-features) only this values are supported for `watch` option: *boolean or []*
Behaviour about string values is not even described. (watching "./true" dir is not strange??)

Anyway feel free to close this PR if is unrelated ;)

soyuka:
It's valid but it's not a boolean. Indeed, a string is specifying a directory to watch and we will not be able to change the behaviour for "true" (because then we would have to do it for "false"). Also, there are other boolean flags, which would not respect this behaviour, which leads to lack of consistency. 

Please use `true` as a boolean if you don't mind ;).

About the PR I'm waiting for other contributors comments on this.

vmarchaud:
People might considering `"true"` as a valid json boolean so i'm okay to merge it (after handling `false` too), indeed we loose consistency but we gain simplicity for end-user.

pi0:
@vmarchaud @soyuka I've changed the casting method:
```js
   if (!app.watch && opts.watch)
      app.watch = JSON.parse(opts.watch);
```
doing `JSON.parse` against on opts.watch, will ensure 3 conditions:
- primitive `true`/`false` => `boolean`
- `"true"`/`"false"` => `boolean`
- other strings => unaccepted (as docs mention only boolean OR arrays allowed)
- [path array] => `array`

Anyway tests are failing as some of them using undocumented form of watch opt (string literal) ! I think it might be better to do following:

1- Mention string support in docs
2- During initialization check for "true"/"false" strings and warn user that she might be doing something wrong. much user friendly and consistency...

PR #2603, opened 2016-12-27
Fix FreeBSD rc.conf overwrite

manuelkasper:
Overwriting rc.conf on FreeBSD is very serious and will lead to a non-working system at the next reboot, as all IP address information, daemon configuration, etc. is lost.

| Q             | A
| ------------- | ---
| Bug fix?      | yes


Unitech:
pm2@2.2.3 published - thx for this PR

PR #2647, opened 2017-01-20
Update README.md

### LINKED ISSUE #2006: pm2 startup not working after updating Node
Hi,

I've pm2 set up with a few processes, that have been saved, then ran `pm2 startup ubuntu`.
It was all working fine: on every server reboot pm2 would restart the processes.

Recently though I updated Node and pm2.

Then, I noticed that the pm2 was not starting up anymore on reboot. I'm not sure if this is due to updating the Node version, or pm2 itself.

In either case I think this is highly undesirable, and it should be fixed. If I set up pm2 to automatically start on reboot, I expect this behaviour to hold true after updating Node or pm2.

Could you please check? Thanks!

p.s. Imho a good suspect for this bug is that the Node executable path changes after a Node update

pensierinmusica:
Add instructions for daemon process when changing Node version.

fixes #2006

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2006
| License       | MIT

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,8 @@ To save/freeze a process list on reboot:
 $ pm2 save
 ```
 
+When you update NodeJS, or switch to another version, you need to rerun `pm2 startup` for the daemon process to work correctly.
+
 [More about startup scripts](http://pm2.keymetrics.io/docs/usage/startup/)
 
 ## Module system


soyuka:
Hmm, this is only because we hard code the `pm2` path in the startup files and that in this case it's located in the `.nvm` directory which changes on update.

The best for this would be to make sure that the `root` user has the `pm2` binary in it's `$PATH`. 

> When you update NodeJS, or switch to another version, you need to rerun `pm2 startup` for the daemon process to work correctly.

I'm suggesting something like this:

> If you use `nvm`, the `pm2` binary may have change after a nodejs update. Therefore, you will need to run `pm2 startup` again for it to be updated. 

This should also be present in the official docs.

vmarchaud:
I updated the docs here : http://pm2.keymetrics.io/docs/usage/startup/#command, so i don't think its necessary to add it into the readme.

soyuka:
+1 thx for reporting @pensierinmusica

pensierinmusica:
@vmarchaud I think this is actually a very important piece of information, and it could be easily overlooked, so it should be in the readme. So many developers are using nvm to manage Node installations.

PR #2650, opened 2017-01-22
Allow starting processes for each environment

jaibatrik:

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | http://stackoverflow.com/questions/29923275/using-pm2-how-can-i-deploy-my-node-js-app-to-multiple-environments-and-ports-on
| License       | MIT

In `ecosystem.config.js`, the names of `apps` are given as hardcoded strings. Thus, it is impossible to start multiple instances of the same application in different environment mode. For example, if I run a particular app in production mode, and simultaneously want to start another instance of the app in the same server (but different port) in development mode, I cannot do that. As soon as I start the app in dev mode, the prod process is replaced by the dev process.

Example JSON where I face the problem - 

```javascript
    apps: [
        // First application
        {
            name: "web",
            script: "index.js",
            env: {},
            env_dev: {
                NODE_ENV: "dev",
                PORT: 4000
            },
            env_prod: {
                NODE_ENV: "prod",
                PORT: 3000
            }
        }
    ]
```

This PR tries to solve it by adding a new property to the app config (`envSpecificInstance`: true|false) and starting the new process with the environment identifier appended to the process name.

With this `pm2 start ecosystem.config.js --env prod` will start process with name `web-prod` and `pm2 start ecosystem.config.js --env dev` with name `web-dev`.

### DIFF
--- a/lib/API.js
+++ b/lib/API.js
@@ -895,7 +895,7 @@ API.prototype._startJson = function(file, opts, action, pipe, cb) {
       app.ignore_watch = opts.ignore_watch;
     if (opts.instances && typeof(opts.instances) === 'number')
       app.instances = opts.instances;
-    if (app.envSpecificInstance) {
+    if (app.envSpecificInstance && opts.env) {
       app.name += ('-' + opts.env);
     }
     apps_name.push(app.name);
--- a/lib/API.js
+++ b/lib/API.js
@@ -895,6 +895,9 @@ API.prototype._startJson = function(file, opts, action, pipe, cb) {
       app.ignore_watch = opts.ignore_watch;
     if (opts.instances && typeof(opts.instances) === 'number')
       app.instances = opts.instances;
+    if (app.envSpecificInstance) {
+      app.name += ('-' + opts.env);
+    }
     apps_name.push(app.name);
   });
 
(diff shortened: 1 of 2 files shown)


vmarchaud:
I think we can maybe find an better name for this option, like `namePerEnv` or `appendEnvToName` ?

jaibatrik:
Thanks for the suggestions, @vmarchaud . I think appendEnvToName makes the purpose clearer. I'll make the change.

jaibatrik:
The failures look unrelated, closing and reopening the PR to trigger the build and check.

jaibatrik:
Verified. The failures are unrelated. They are failing because of https://github.com/Unitech/pm2/commit/af3d115edc9b3562ba81a24d01831981428ce209.

vmarchaud:
Totally unrelated, i restarted the build and it should pass now.

vmarchaud:
It will land on master in February, thanks for your contribution.

vmarchaud:
Just to update you, we changed the option name to use `snake_case` since all options are using it right now. see https://github.com/Unitech/pm2/commit/6f5ddaf7aee02547a1faabce1bfeb957e89a84b0

jaibatrik:
Yeah, should have noticed that. Thanks!

szalishchuk:
This happened to be useful for log discovery with Kibana, thank you. But, of course, it would be better to have env_name as a separate json property in every log entry.

RobertShaw1:
Doesn't look like this option ever made its way into the documentation.  How can I contribute?

Issue #2648, opened 2017-01-20
Can pm2 ecosystem support promise config


kimown:
Hi, I am using `pm2 start ecosystem.config.js` to start my multiple express projects, I use [detect-port](https://github.com/node-modules/detect-port) in each express project to avoid using used ports. Because free port is a dynamic number, so I want to show port in App name when I run `pm2 list`
```
┌──────────────┬────┬──────┬───────┬────────┬─────────┬────────┬─────┬────────┬──────────┐
│ App name     │ id │ mode │ pid   │ status │ restart │ uptime │ cpu │ mem    │ watching │
├──────────────┼────┼──────┼───────┼────────┼─────────┼────────┼─────┼────────┼──────────┤
│ script       │ 0  │ fork │ 33473 │ online │ 4       │ 6h     │ 0%  │ 0 B    │ disabled │
│ specify port │ 12 │ fork │ 35716 │ online │ 0       │ 2h     │ 0%  │ 0 B    │ disabled │
└──────────────┴────┴──────┴───────┴────────┴─────────┴────────┴─────┴────────┴──────────┘
```
I searched some issues, the only way I found doing this is change the name before run `pm2 start ecosystem.config.js`.
```
const apps = [
    {
        name:'specify port',
        script    : "app.js",
    },
    {
        name:'script',
        script    : "web.js"
    }
];
module.exports = apps;
```

So if pm2 can support promise config, my code can look like this:
```
const detect = require('detect-port');

function getPorts(basePort, number) {
    let ports = [];
    return new Promise((resolve)=>{
        const cb = (err, _port) => {
            if (err) {
                console.log(err);
            }
            ports.push(_port)
            if(ports.length<number){
                detect(_port+1, cb)
            }else {
                resolve(ports)
            }
        }

        detect(basePort, cb);
    })
}

const apps = [
    {
        name:'specify port',
        script    : "app.js",
    },
    {
        name:'script',
        script    : "web.js"
    }
];


function getApp() {
    return new Promise((resolve)=>{
        getPorts(3000, apps.length).then((data)=>{
            const appCopy = apps.map((v,k)=>{
                return Object.assign(v,{name:v.name+data[k]})
            })
            resolve(appCopy)
        })
    })
}

module.exports = getApp
```
Or can you give me other suggestions, thanks.

soyuka:
Would be a nice feature IMO!

kimown:
If the PR merged, you can run `pm2 start ecosystem.config.js` like this:

[...]

PR #2658, opened 2017-01-24
(features) support promise in process file 

kimown:

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2648
| License       | MIT


lib/API.js
```diff
@@ -836,141 +836,151 @@ API.prototype._startJson = function(file, opts, action, pipe, cb) {
...
-          // And Remove from array to spy
-          apps_name.splice(apps_name.indexOf(proc_name), 1);
-          return next();
-        });
-      });
+        try {
+          data = fs.readFileSync(file_path);
+        } catch(e) {
+          Common.printError(conf.PREFIX_MSG_ERR + 'File ' + file +' not found');
+          return cb ? cb(Common.retErr(e)) : that.exitCli(conf.ERROR_EXIT);
```
soyuka: Does this `cb` still exists?

lib/API.js
```diff
@@ -836,141 +836,151 @@ API.prototype._startJson = function(file, opts, action, pipe, cb) {
...
-    return false;
-  });
+        try {
+          Common.parseConfig(data, file).then(function (config) {
+            resolve(config)
+          });
+        } catch(e) {
+          Common.printError(conf.PREFIX_MSG_ERR + 'File ' + file + ' malformated');
+          console.error(e);
+          return cb ? cb(Common.retErr(e)) : that.exitCli(conf.ERROR_EXIT);
```
soyuka: Same here?

lib/API.js
```diff
@@ -1046,109 +1056,120 @@ API.prototype._startJson = function(file, opts, action, pipe, cb) {
...
+
+        try {
+          Common.parseConfig(data, file).then(function (appConf) {
+            resolve(appConf)
+          });
+        } catch(e) {
+          Common.printError(conf.PREFIX_MSG_ERR + 'File ' + file + ' malformated');
+          console.error(e);
+          return cb ? cb(Common.retErr(e)) : that.exitCli(conf.ERROR_EXIT);
+        }
```
soyuka: If `Common.parseConfig` is a Promise, we should leverage the use of promise:

```
//the then part is not necessary if you resolve appConf 
return Common.parseConfig()
//but you should catch it to display that proper message
.catch(function(err) {
  Common.printError(conf.PREFIX_MSG_ERR + 'File ' + file + ' malformated');
  console.error(e);
  return Promise.reject(Common.retErr(e)) //not sure that `retErr` function has to be used
})

File: lib/API.js [hunk 4]
```diff
@@ -1046,109 +1056,120 @@ API.prototype._startJson = function(file, opts, action, pipe, cb) {
...
+      } else if (jsonVia == 'pipe') {
+        Common.parseConfig(file, 'pipe').then(function (appConf) {
+          resolve
```
soyuka: wat

lib/API.js
```diff
@@ -1046,109 +1056,120 @@ API.prototype._startJson = function(file, opts, action, pipe, cb) {
...
+      else return that.speedList();
     });
-  }, function(err) {
-    if (cb) return cb(null, ret_processes);
-    else return that.speedList();
-  });
+  })
+
+  function getAppConf() {
+    return new Promise(function (resolve) {
```
soyuka: Please try to avoid `new Promise`, instead just do `Promise.resolve()` (if you can, looks like it's the case here).

lib/API/Version.js
```diff
@@ -306,32 +306,32 @@ module.exports = function(CLI) {
...
+                  }
+                  else
+                    return callb();
+                }, function(final) {
+                  return callback(final);
+                });
+              }
+              else {
+                return callback();
+              }
```
soyuka: What about:

```
if (!data || !data.apps) {
 return callback()
}

//rest of the code
```

lib/Common.js
```diff
@@ -242,30 +243,39 @@ Common.parseConfig = function(confObj, filename) {
...
-  }
+    else if (filename.indexOf('.yml') > -1 ||
+      filename.indexOf('.yaml') > -1) {
+      resolve(yamljs.parse(confObj.toString()));
+    }
+    else if (filename.indexOf('.config.js') > -1) {
+      var confPath = require.resolve(path.resolve(filename));
+      delete require.cache[confPath];
+      // Probably a promise
+      if (typeof require(confPath).then === 'function') {
```
soyuka: This is not sufficient and you `require` the item twice

```
var conf = require(confPath)

if (typeof conf === 'object' && conf.then === 'function') {
  conf.then()...
}
```

soyuka:
Nice one! Sorry I should've used the `review` function!

I made a few comments, it's a good start!

lib/API.js
vmarchaud: Used if you use the programmatic API

lib/API.js
```diff
@@ -836,141 +836,151 @@ API.prototype._startJson = function(file, opts, action, pipe, cb) {
...
 
-  if (typeof(file) === 'object')
-    config = file;
-  else if (pipe == 'pipe')
-    config = Common.parseConfig(file, 'pipe');
-  else {
-    var data = null;
-
-    var file_path = path.join(that.cwd, file);
+  getConfig().then(function (config) {
```
vmarchaud: Need to use `reject()` and catch it here to call `cb` or exit the CLI with the error.

lib/API.js
kimown: @soyuka Yes, but if I use this way, I must add `return Promise.resolve(xxx)` or `return Promise.resolve(xxx)` in every return statement, I think a function only have one return statement is more clear. Also, I don't understand why avoid `new Promise `, In my project, I almost use `new Promise` everywhere when I need to use promise, is it wrong to use `new Promise`?

vmarchaud:
I think native Promise aren't fully available on node `0.12` so this PR may need to wait until we drop the support for this nodejs version.
In the mean time, you can some basic integration test (just to verify that the paste you provided in https://github.com/Unitech/pm2/issues/2648 is correctly resolved), like this we will see if Promise are supported in node 0.12 or not.

vmarchaud:
We have frozen feature for pm2 `2.4.0` which will be published monday, but this PR will definitely be in the next iteration !

kimown:
@vmarchaud I see, but one job in Travis CI fails, can you help me restart the job

vmarchaud:
@kimown Restarted the jobs and it passing, we should merge this with pm2 `2.4` if everything is good !

kimown:
@vmarchaud ok, I also update the demo https://github.com/Unitech/pm2/issues/2648#issuecomment-278843809

cKehres:
can some fix it so that anyone can use it? ;) i would start my ecosystem based on database entries, so promise support would be very nice

PR #2770, opened 2017-03-15
Fix trigger API never calling callback

yaroslav-korotaev:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none
| License       | MIT

When using PM2 as API, sometimes it doesn't call the callback passed to trigger() function. I've found that it happens when msgProcess()'s callback is called before the bus event is received (process_wait_count still be a zero). So, we wait for bus launch, register listener on it, and then send a message to process.

### DIFF
--- a/lib/API/Extra.js
+++ b/lib/API/Extra.js
@@ -163,16 +163,16 @@ module.exports = function(CLI) {
             return cb ? cb(null, results) : that.exitCli(cst.SUCCESS_EXIT);
         }
       });
-    });
 
-    that.msgProcess(cmd, function(err, data) {
-      if (err) {
-        Common.printError(err);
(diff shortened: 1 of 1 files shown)


vmarchaud:
Good catch, i'll merge when the tests pass.

yaroslav-korotaev:
I've repeated local test launch with something like this:
```
cd ~/dev/pm2
bin/pm2 kill
cd ~/dev
rm -rf pm2
rm -rf ~/.pm2
rm -rf ~/.pm2-dev
git clone git@github.com:yaroslav-korotaev/pm2.git
cd pm2
git checkout development
npm install
npm test
```

And it's pass.
Ubuntu 16.04, Node 6.10.0, nvm and php installed.

PR #2810, opened 2017-04-03
Fix: Callbacks for process.stdout / .stderr

rluba:
The overrides for `process.stdout.write` and `process.stderr.write` did not invoke the provided callbacks when they’re done.

This solution is a bit hacky because it just waits for `stds.out` and `stds.err`, respectively. A proper fix should wait for all of the downstream writes (`std.std` and/or `std.out/err` and/or `process.send`, depending on the configuration), before invoking the callback – and forward any potential errors.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no, but the same four tests also fail on `development`
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

### DIFF
--- a/lib/ProcessContainer.js
+++ b/lib/ProcessContainer.js
@@ -204,9 +204,12 @@ function exec(script, stds) {
         stds.std && stds.std.write && stds.std.write(log_data);
 
         // hardcoded values of special log path to not write on disk
-        if (pm2_env.pm_err_log_path !== 'NULL' && pm2_env.pm_err_log_path !== '/dev/null') {
-          stds.err.write && stds.err.write(log_data, encoding, cb);
-        };
+        if (pm2_env.pm_err_log_path !== 'NULL' && pm2_env.pm_err_log_path !== '/dev/null' && stds.err.write) {
+          stds.err.write(log_data, encoding, cb);
+        }
+        else {
+          cb();
+        }
 
         process.send({
           type : 'log:err',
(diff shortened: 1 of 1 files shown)


lib/ProcessContainer.js
```diff
@@ -207,7 +207,7 @@ function exec(script, stds) {
         if (pm2_env.pm_err_log_path !== 'NULL' && pm2_env.pm_err_log_path !== '/dev/null' && stds.err.write) {
           stds.err.write(log_data, encoding, cb);
         }
-        else {
+        else if (cb) {
```
soyuka: cb !== undefined

rluba: Sure, I can change it, if you prefer that version. But the whole codebase seems to use boolean coercion all the time – just as I did.

Some examples:

```JavaScript
if (process.connected &&
    process.send &&
    process.versions &&
    process.versions.node)
        …
  });
```

or 

```JavaScript
if (… && stds.out.write) {
    …
}
```

soyuka: yeah I know, just my 2 cents :). The PR looks fine to me anyway!

lib/ProcessContainer.js
```diff
@@ -238,8 +241,11 @@ function exec(script, stds) {
         stds.std && stds.std.write && stds.std.write(log_data);
```
vmarchaud: Should we change this to `write(log_data, encoding)` here too ?

soyuka: we definitely should, nice catch!

rluba: Thanks, done.

PR #2816, opened 2017-04-06
Fix #2815 nodejs 0.11 compatibility is Absolute

### LINKED ISSUE #2815: Object #<Object> has no method 'isAbsolute'
#2797 is not compatible with Node version below `0.11.2` as `path.isAbsolute` was only introduced in  `0.11.2`.

soyuka:

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2815
| License       | MIT

### DIFF
--- a/lib/API.js
+++ b/lib/API.js
@@ -833,17 +833,27 @@ API.prototype._startJson = function(file, opts, action, pipe, cb) {
   var apps_info  = [];
   var that = this;
 
-  if (typeof(cb) === 'undefined' && typeof(pipe) === 'function')
+  if (typeof(cb) === 'undefined' && typeof(pipe) === 'function') {
     cb = pipe;
+  }
 
-  if (typeof(file) === 'object')
+  if (typeof(file) === 'object') {
     config = file;
-  else if (pipe == 'pipe')
+  } else if (pipe === 'pipe') {
     config = Common.parseConfig(file, 'pipe');
-  else {
+  } else {
     var data = null;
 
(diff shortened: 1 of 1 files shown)


vmarchaud:
We dropped support for nodejs version under 0.12 so i guess we shouldnt merge this one @Unitech  ?

soyuka:
Fine by me \o/. But If you do dropped the support we need to update [the package.json](https://github.com/Unitech/pm2/blob/master/package.json#L6) `engines`.

vmarchaud:
Lets merge the PR and changes the required engine in the package.json so

soyuka:
@vmarchaud WDYT about the engine, should we set it to LTS? Or maybe` >=4`? I'd personally go with `4` instead of `0.12`

vmarchaud:
@soyuka We discussed to drop the support for 0.12 too when we dropped the 0.10 (beginning january), but @Unitech wanted to keep it for a little longer, maybe now should be a good time ? (drop a the same time both)

Unitech:
It's ok to merge that PR for now.
We will drop support for node < 4 on PM2 v3

vmarchaud:
@Unitech But do we upgrade the engine needed to `>=0.12` in the `package.json` like @soyuka said (that will forbid update with node 0.10) ?

Unitech:
yes let's do this

PR #2876, opened 2017-05-11
(bugfix) #2868: fix --raw not display previous logs

### LINKED ISSUE #2868: pm2 logs --raw doesn't display previous logs
I'm trying to display my app logs with PM2, and they're in JSON, so I need raw logs. 
When using the `pm2 logs <app-name> --raw`, there are only logs that were emitter after the command start, the previous logs aren't displayed (though if I run it without the `--raw` flag, the previous logs are displayed).
The `--lines <n>` or `--lines=<n>` options seem to not work.

I've seen this issue: https://github.com/Unitech/PM2/issues/1277, however it was stated that it has been fixed and I'm still facing it.

Can you help me with it?

```
PM2 version: `2.4.6`
Node version: `7.10.0`, installed with `n`
Xubuntu Studio 17.04 x64
```

poppinlp:
Fixed #2868 : now will display previous logs for raw option.

### DIFF
--- a/lib/API/LogManagement.js
+++ b/lib/API/LogManagement.js
@@ -143,7 +143,7 @@ module.exports = function(CLI) {
         that.exitCli(cst.ERROR_EXIT);
       }
 
-      if (lines === 0 || raw)
+      if (lines === 0)
         return Log.stream(that.Client, id, raw, timestamp, exclusive);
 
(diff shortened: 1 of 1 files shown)


poppinlp:
The condition which i modified leads just `Log.stream` for `raw` config all time.
Remove `or` for `raw` will go to `Log.tail` bellow and display previous logs.
I don't understand what's that condition for. So you may check for that.
And sorry for bother you if i made a wrong PR.

PR #2886, opened 2017-05-17
small wording fix

kedadi:
without pmx -> without automation


vmarchaud:
`--no-automation`[ is used](https://github.com/Unitech/pm2/blob/1c3fcb83b65e430a5734d576610d27cc7a0bf4fd/lib/Common.js#L91) to disable pmx.

PR #2892, opened 2017-05-21
fixing the generated sudo script for non root users

bitcloud:
the original script evaluated the $PATH variable still in the user context which in some distriburions doesn't include the sbin folders. I changed it so it will be evaluated in the target (root) user context.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2892
| License       | MIT
| Doc PR        | na


bitcloud:
this was now already fixed in upstream.

Issue #452, opened 2014-05-09
HttpInterface exposes critical environment variables


MichielDeMey:
Hi,

My issue is that currently, the HttpInterface API exposes critical environment variables such as usernames and passwords.

I would suggest having an option to strip environment variables and other potentially sensitive information from the API, since we don't always want to expose this to others.

jrottenberg:
The pull request is not valid anymore (need a rebase), but is there anything beside that prevents the integration ?

LukasBeaton:
Any plans to merge this in?

This would be really nice to have!!!

PR #2895, opened 2017-05-23
Stripping environment variables from the HttpInterface

MichielDeMey:

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #452
| License       | MIT

Re-submit of https://github.com/Unitech/pm2/pull/459

Let's try not to wait 3 years this time. 😉

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -80,7 +80,7 @@ var csts = {
   DEBUG                   : process.env.PM2_DEBUG || false,
   WEB_IPADDR              : process.env.PM2_API_IPADDR || '0.0.0.0',
   WEB_PORT                : parseInt(process.env.PM2_API_PORT)  || 9615,
-  WEB_STRIP_ENV_VARS      : process.env.PM2_WEB_STRIP_ENV_VARS || true,
+  WEB_STRIP_ENV_VARS      : process.env.PM2_WEB_STRIP_ENV_VARS || false,
   MODIFY_REQUIRE          : process.env.PM2_MODIFY_REQUIRE || false,
 
   WORKER_INTERVAL         : process.env.PM2_WORKER_INTERVAL || 30000,
--- a/constants.js
+++ b/constants.js
@@ -80,6 +80,7 @@ var csts = {
   DEBUG                   : process.env.PM2_DEBUG || false,
   WEB_IPADDR              : process.env.PM2_API_IPADDR || '0.0.0.0',
   WEB_PORT                : parseInt(process.env.PM2_API_PORT)  || 9615,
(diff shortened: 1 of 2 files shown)


lib/HttpInterface.js
```diff
@@ -46,6 +46,17 @@ function startWebServer(pm2) {
           processes: list
         };
 
+        if (cst.WEB_STRIP_ENV_VARS === true) {
+          for (var i = data.processes.length - 1; i >= 0; i--) {
+            var process = data.processes[i];
```
soyuka: can you use another variable name? This is a global nodejs variable.

MichielDeMey: Good point, I'll use `proc`.

constants.js
```diff
@@ -80,6 +80,7 @@ var csts = {
   DEBUG                   : process.env.PM2_DEBUG || false,
   WEB_IPADDR              : process.env.PM2_API_IPADDR || '0.0.0.0',
   WEB_PORT                : parseInt(process.env.PM2_API_PORT)  || 9615,
+  WEB_STRIP_ENV_VARS      : process.env.PM2_WEB_STRIP_ENV_VARS || true,
```
soyuka: I'm not sure about the default value for this, might break some implementations. WDYT @vmarchaud, shouldn't we set this to `false` by default so that things remain unchanged?

MichielDeMey: Not sure about some internal APIs, but yeah this might introduce a breaking change. I simply opted for the "secure by default" option, but we might want to set this to `false` in case it does break backwards compatibility.

vmarchaud: @soyuka The behavior shouldn't be changed yes, by default we should put it as false.

MichielDeMey: Should I rename the ENV var then as well? Suggestions welcome. :)

soyuka: name is fine to me!

MichielDeMey:
CI seems to fail on Node 6, known issue?

soyuka:
Nothing related imo.

lib/HttpInterface.js
```diff
@@ -46,6 +46,17 @@ function startWebServer(pm2) {
           processes: list
         };
 
+        if (cst.WEB_STRIP_ENV_VARS === true) {
+          for (var i = data.processes.length - 1; i >= 0; i--) {
+            var proc = data.processes[i];
+
+            // Strip important environment variables
+            if (typeof proc.pm2_env === 'undefined' && typeof proc.pm2_env.env === 'undefined') return;
+
```
susan51531: Maybe change 'return' to 'break;

http request will not be ended if return from here.

PR #2904, opened 2017-05-25
add wait ready for cluster mode

evollu:
fixes ##2890


evollu:
looks like `should trigger message by name` test failed for fork mode but my only change is for cluster mode. any idea what is happening?
@vmarchaud

vmarchaud:
@evollu Some test are really flaky so its normal that sometime it fail for no reason. If it pass for all old versions it should be good.

PR #3182, opened 2017-09-29
force upgrade to latest pm2-deploy, fixes #3181

### LINKED ISSUE #3181: force upgrade of pm2-deploy
https://github.com/Unitech/pm2-deploy/issues/121 is still causing problems to user because the semver range is still valid. imho is better to force the upgrade to the last version.

phra:
see #3181


Unitech:
could you make this PR on #development? Thanks

PR #3193, opened 2017-10-04
Retain multiple newlines in log stream

shanehughes3:
Currently, if any process tries to log a blank line (i.e. `\n\n`), the line is removed in streaming logs yet retained in the log file and in the tail function. This PR retains those extra newlines.

The Log.*stream methods are checking for general falsiness of each element after splitting a chunk on `\n`, presumably to remove the resulting last element that will be an empty string. We can double-check that an element is indeed the last element before ignoring it.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT

### DIFF
--- a/lib/API/Log.js
+++ b/lib/API/Log.js
@@ -113,8 +113,8 @@ Log.stream = function(Client, id, raw, timestamp, exclusive) {
       else
         return;
 
-      lines.forEach(function(line) {
-        if (!line) return;
+      lines.forEach(function(line, i) {
+        if (!line && i === lines.length - 1) return;
 
         if (raw)
           return process.stdout.write(util.format(line) + '\n');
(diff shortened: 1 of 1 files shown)


vmarchaud:
Thanks for bringing this problem, although your fix can produce unwanted behavior (maybe that the last line is a message in the case without two \n), so based on your work i made https://github.com/Unitech/pm2/commit/4f5566ea93dd6577d12a84c5b3ac610f10acea76 that should fix the issue :)

shanehughes3:
4f5566e does not fix the issue - see #3237.

Issue #3186, opened 2017-10-02
Request: TypeScript Definition Files


nicky-lenaers:
Your issue may already be reported!
Please search on the [issue tracker](https://github.com/Unitech/pm2/search?type=Issues) before creating one.

## What's going wrong?
Can't find any TypeScript Definition Files to use `pm2` programmatically.

## How could we reproduce this issue?
Use `pm2` programmatically with TypeScript using TypeScript definition files for strict type checking. A simple example for the `connect` function might be:

```ts
declare module 'pm2' {
  export type ConnectCallback = (err?: Error) => void;
  export function connect(cb: ConnectCallback): void;
}
```

## Supporting information

Please run the following command (available on PM2 >= 2.6)

```
$ pm2 report
```
```
===============================================================================
--- PM2 REPORT (Mon Oct 02 2017 10:13:38 GMT+0200 (W. Europe Daylight Time)) --
===============================================================================
--- Daemon -------------------------------------------------
pm2d version         : 2.7.0
node version         : 8.2.1
node path            : C:/Program Files/nodejs/node.exe
argv                 : C:\Program Files\nodejs\node.exe,C:\Users\<user_dir>\<project_dir>\node_modules\pm2\lib\Daemon.js
argv0                : node
user                 : undefined
uid                  : N/A
gid                  : N/A
uptime               : 729min
===============================================================================
```

vmarchaud:
@jportela did a PR for this one, thanks again :)

PR #3202, opened 2017-10-08
Add first draft of typescript definitions

jportela:
I've created a first draft of the typescript definitions file. I've attached a quick gif showing how it helps on developer workflow, even if the developer is not using Typescript (baked in in Visual Studio Code, but can easily be setup for other editors).

I've relied on the http://pm2.keymetrics.io/docs/usage/pm2-api/ documentation to create the definitions, since there's a lot of comments there I could reuse. However, this might not reflect accurately what's in code, so I'll be making a second pass at the code, making sure it accurately translates what's on the documentation. I might also add other functions that are not documented.

Let me know if this is a good direction to take this into, I hope this is helpful.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3186 
| License       | MIT

typescript-definitions

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
   "homepage": "http://pm2.keymetrics.io/",
   "description": "Production process manager for Node.JS applications with a built-in load balancer.",
   "main": "index.js",
+  "types": "types/index.d.ts",
   "scripts": {
     "test": "NODE_ENV=test bash test/pm2_programmatic_tests.sh && NODE_ENV=test bash test/pm2_behavior_tests.sh",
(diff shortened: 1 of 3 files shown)


vmarchaud:
Thanks for looking into this, it definitely is the right direction.
I will try to add few myself and correct some if there is any error :)

jportela:
as a reference, here's the generated typescript definitions by dts-gen:
https://gist.github.com/jportela/7e22bcd12375d3144e30d4445a144d46

PR #3213, opened 2017-10-17
Allow custom service names for startup scripts

danez:
| Q             | A
| ------------- | ---
| Bug fix?      | 
| New feature?  | y
| BC breaks?    | 
| Deprecations? | 
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Especially when running mutliple instances of pm2 on one machine with
the same user this helps to create multiple startup scripts for each of
the instance.


danez:
Test failure seems unrelated, at least I do not see what is failing.

the startup test isn't even run as it needs root

vmarchaud:
Indeed can be really useful, thanks for taking the time to make this PR too

PR #3223, opened 2017-10-20
Allow installing modules from local file destinations

danez:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

For example `pm2 install file:///home/user/module`

### DIFF
--- a/lib/Utility.js
+++ b/lib/Utility.js
@@ -213,6 +213,11 @@ var Utility = module.exports = {
     if(canonic_module_name.indexOf('http') !== -1) {
       canonic_module_name = canonic_module_name.split('/').pop();
     }
+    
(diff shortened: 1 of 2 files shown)


danez:
We use this for production. We want to be in control of all the dependencies so we install `pm2` and `pms-logrotate` in the project and then do `pm2 install node_modules/pm2-logrotate`, that way we can control which version of `pm2-logrotate` is used.

vmarchaud:
Okay i see your use case ! Indeed, its easier to control the versioning directly with the package.json :)

PR #3226, opened 2017-10-23
Fix applying env PM2_CONCURRENT_ACTIONS correctly

### LINKED ISSUE #3036: concurrent_actions is hardcoded to 2 on node >= 4
## What's going wrong?
Even after setting `PM2_CONCURRENT_ACTIONS` to a higher value, its being overridden to 2 here: https://github.com/Unitech/pm2/blob/fe4d600b0a9a9c1474b5154bf766471cc48e4b47/constants.js#L77

## How could we reproduce this issue?
Try to set `PM2_CONCURRENT_ACTIONS` to a higher value than 2

## Supporting information

pm2 2.4.6

danez:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | Fixes #3036
| License       | MIT
| Doc PR        |

Now the default will be either 1 or 2 depending on node version and the environment variable will still be able to overwrite the default.

I also tried the --parallel flag first, but neither in the json config nor in the cli it works with `pm2 reload`

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -72,9 +72,10 @@ var csts = {
 
   // Concurrent actions when doing start/restart/reload
   CONCURRENT_ACTIONS      : (function() {
-    var concurrent_actions = parseInt(process.env.PM2_CONCURRENT_ACTIONS) || 1;
+    var default_concurrent_actions = 1;
     if (semver.satisfies(process.versions.node, '>= 4.0.0'))
-      concurrent_actions = 2;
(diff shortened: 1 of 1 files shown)


vmarchaud:
Good catch, indeed we were overriding the number on `node > 4`. Thanks for take the time to make a PR :)

Issue #3206, opened 2017-10-10
pm2-docker in watch and restart mode causes pm2 to exit sometimes


dguo:
## What's going wrong?
For local development, I work inside a Docker container. When I use pm2-docker with watch mode on, pm2 frequently quits after I change a file. I get the following trace:
```
0 application online, exiting
Exiting PM2
```

I did some debugging and found [this commit](https://github.com/Unitech/pm2/commit/4f8b8d6803ab3b626c7da567a093870ee10ab2bc). My guess is that there is a race condition between the setInterval in the autoExit function and the application restarting. If it doesn't restart fast enough, autoExit is triggered.

## How could we reproduce this issue?
Try pm2-docker with watch mode on, and change a file repeatedly.

## Supporting information
I saw the rationale for making autoExit the default for pm2-docker, but [the option](https://github.com/Unitech/pm2/blob/master/lib/binaries/Runtime4Docker.js#L27) is still in the CLI, even though the code [doesn't seem to check for it](https://github.com/Unitech/pm2/blob/master/lib/binaries/Runtime4Docker.js#L43). I think either the flag should be used to turn off autoExit, or it should be removed from the CLI if it doesn't actually do anything.

In the meantime, I switched to using pm2-dev, which is working so far.

```
===============================================================================
--- PM2 REPORT (Tue Oct 10 2017 20:40:41 GMT+0000 (UTC)) ----------------------
===============================================================================
--- Daemon -------------------------------------------------
pm2d version         : 2.7.1
node version         : 8.6.0
node path            : undefined
argv                 : /usr/local/bin/node,/app/node_modules/pm2/lib/Daemon.js
argv0                : node
user                 : undefined
uid                  : 0
gid                  : 0
uptime               : 0min
===============================================================================
--- CLI ----------------------------------------------------
local pm2            : 2.7.1
node version         : 8.6.0
node path            : undefined
argv                 : /usr/local/bin/node,/app/node_modules/.bin/pm2,report
argv0                : node
user                 : undefined
uid                  : 0
gid                  : 0
===============================================================================
--- System info --------------------------------------------
arch                 : x64
platform             : linux
type                 : Linux
cpus                 : Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
cpus nb              : 2
freemem              : 1549955072
totalmem             : 2096160768
home                 : /root
===============================================================================
--- PM2 list -----------------------------------------------
[...]

PR #3232, opened 2017-10-24
Add flag to turn off auto exit with pm2-docker

dguo:
Because we want auto exit to be on by default for pm2-docker, the flag needs to be changed to explicitly turn it off. See the example at https://tj.github.io/commander.js/#Command.prototype.option for a "simple boolean defaulting to true".

The suggested fix was:
```js
if (commander.autoExit !== false) {
 autoExit();
}
```
But I couldn't find a way to actually pass "false" as a value from the CLI.

If you would like to add a test for this change, l could use some help on that.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no (print a warning message if --auto-exit is provided though)
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3206 
| License       | MIT
| Doc PR        | https://github.com/pm2-hive/pm2-hive.github.io/pull/116

### DIFF
--- a/lib/binaries/Runtime4Docker.js
+++ b/lib/binaries/Runtime4Docker.js
@@ -24,12 +24,12 @@ commander.version(pkg.version)
   .option('--delay <seconds>', 'delay start of configuration file by <seconds>', 0)
   .option('--web [port]', 'launch process web api on [port] (default to 9615)')
   .option('--only <application-name>', 'only act on one application of configuration')
-  .option('--auto-exit', 'exit if all processes are errored/stopped or 0 apps launched', true)
+  .option('--no-auto-exit', 'do not exit if all processes are errored/stopped or 0 apps launched')
   .option('--env [name]', 'inject env_[name] env variables in process config file')
   .option('--watch', 'watch and restart application on file change')
   .option('--error <path>', 'error log file destination (default disabled)', '/dev/null')
   .option('--output <path>', 'output log file destination (default disabled)', '/dev/null')
-  .usage('pm2-docker app.js');
+  .usage('app.js');
 
 function start(cmd, opts) {
   pm2 = new PM2.custom({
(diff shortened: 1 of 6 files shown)


lib/binaries/Runtime4Docker.js
```diff
@@ -24,12 +24,12 @@ commander.version(pkg.version)
   .option('--delay <seconds>', 'delay start of configuration file by <seconds>', 0)
   .option('--web [port]', 'launch process web api on [port] (default to 9615)')
   .option('--only <application-name>', 'only act on one application of configuration')
-  .option('--auto-exit', 'exit if all processes are errored/stopped or 0 apps launched', true)
+  .option('--no-auto-exit', 'do not exit if all processes are errored/stopped or 0 apps launched')
   .option('--env [name]', 'inject env_[name] env variables in process config file')
   .option('--watch', 'watch and restart application on file change')
   .option('--error <path>', 'error log file destination (default disabled)', '/dev/null')
   .option('--output <path>', 'output log file destination (default disabled)', '/dev/null')
-  .usage('pm2-docker app.js');
```
dguo: The help usage was `Usage: pm2-docker pm2-docker app.js`, so this change removes the duplicate `pm2-docker`.

vmarchaud:
I think we can let a fake flag to maintain backward compatibility (cause a lot of people might using this flag), and we could even add a error message when the flag is used, something like : 
 `Warning: --auto-exit has been removed, it's now the default behavior, if you want to disable it a new flag has been introduced : --no-auto-exit.`) 
What do you think ?

dguo:
That makes sense. I added the warning message and rebased. This should be a backwards compatible change now.

Instead of leaving the `--auto-exit` flag in, I added a check to `process.argv`. I tried the fake flag approach, but Commander doesn't seem to handle having both `--foo` and `--no-foo` options. The value would come back as `false`, even when the flag was provided. 

Another approach would be to use Commander's [allowUnknownOption](https://tj.github.io/commander.js/#Command.prototype.allowUnknownOption), but then we would still have to check `process.argv` to provide the warning message, and we would lose the error handling for truly unknown options.

vmarchaud:
That sounds good good to me, thanks for taking the time to make this PR :)

PR #3316, opened 2017-11-28
raw output should keep the original output stream.

serkanyersen:
in docker/kubernetes setup, error logging depends on errors being on stderr. My app does this by default but when PM2 takes over, all logs gets combined into stdout. Which makes it really hard to automatically detect the errors. 

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT


Unitech:
nice catch

Issue #2182, opened 2016-05-21
How to get rid of annoying popups in Windows 10?


Kitanotori:
How do I disable the popup windows that pm2 generates in Windows 10 when starting, restarting or reloading a Node.js app? From the perspective of rapid server-side development, this is very annoying...

Unitech:
Would it be an issue because of the way we spawn processes?

- https://github.com/Unitech/pm2/blob/development/lib/God/ClusterMode.js#L52
- https://github.com/Unitech/pm2/blob/development/lib/God/ForkMode.js#L94

I don't know if I have to customize something on this code segments,

soyuka:
`{detached: true` should get rid of this error but it is not available for cluster.

Unitech:
interesting, maybe @ThisIsMac47 will have a look at this on the next weeks

vmarchaud:
After testing, the detached option doesnt hide popups in windows, i think that nodejs spawn a shell to spawn the process.
I will investigate more in the next days.

vmarchaud:
Okay so : 
- In fork mode, you shouldnt have popups in the dev branch since now
- In cluster mode, unfortunately, we dont have control on the popup cause we use the [cluster implementation of nodejs](https://nodejs.org/dist/latest-v6.x/docs/api/cluster.html).

If you get any popup apart of cluster mode, respond to this issue to report it, i will investigate for each one.

soyuka:
Because we can't get `detached: true` in cluster?

vmarchaud:
They use the function fork, we can use ```detached: true``` only with spawn.

soyuka:
Yes ok so if that's the main reason, the only solution is to port cluster into pm2 instead of using the nodejs cluster. 
Did you try a simple cluster implementation without pm2 to see if it pops up too?

vmarchaud:
I wasnt able to reproduce with a simple cluster implementation, this might coming from the fact that we are a spawning a child from a forked child or something like this.
We manage to delete the popups in cluster mod by [rewriting the implementation](https://github.com/Unitech/pm2/commit/0f2ecdd4b5cfe305a0a58f5033a142c793fbb993) on the [branch of a forked cluster implementation](https://github.com/Unitech/pm2/tree/cluster-embedded), we will talk about how include this into future release of pm2.

ajithr:
Is this issue fixed or any workaround available? I have faced the same issue when using cluster mode with watch option. The node window opens multiple time when saving the file.

[...]

PR #3400, opened 2018-01-12
#2182 Get rid of annoying popups in Windows 10

toddwong:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2182
| License       | MIT

Works with node v9.4.0 and above.

### DIFF
--- a/lib/God/ForkMode.js
+++ b/lib/God/ForkMode.js
@@ -93,7 +93,6 @@ module.exports = function ForkMode(God) {
       try {
         var cspr = spawn(command, args, {
           env      : pm2_env,
-          detached : true,
           windowsHide: true,
           cwd      : pm2_env.pm_cwd || process.cwd(),
           stdio    : ['pipe', 'pipe', 'pipe', 'ipc'] //Same as fork() in node core
--- a/lib/God/ForkMode.js
+++ b/lib/God/ForkMode.js
@@ -93,6 +93,7 @@ module.exports = function ForkMode(God) {
       try {
         var cspr = spawn(command, args, {
           env      : pm2_env,
+          detached : true,
           windowsHide: true,
           cwd      : pm2_env.pm_cwd || process.cwd(),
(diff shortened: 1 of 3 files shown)


lib/God/ForkMode.js
```diff
@@ -93,7 +93,6 @@ module.exports = function ForkMode(God) {
       try {
         var cspr = spawn(command, args, {
           env      : pm2_env,
-          detached : true,
```
soyuka: do you really need to remove detached here?

toddwong: @soyuka I can't make the windowsHide option works with `detached: true`. Do we actually needs detached to be true here? I maybe mistook something here, but child processes in cluster mode are attached to the daemon process I think.

toddwong: @soyuka I tried to reproduce the problem (`detached:true` with 'windowsHide:true`) to find the real cause, but found something unexpected. I think the console windows may be not caused by this spawn call at all, but something else. I'll do some digging.

toddwong:
@soyuka I've found the root cause of the popping up console window while start process in fork mode.  It's not the `spawn` call in lib/God/ForkMode.js, but `pmx.init` in lib/ProcessContainerFork.js. I've submit a PR to the pmx project, and undone the `detached: true` deletion.

soyuka:
Great work @toddwong thanks!

vmarchaud:
Tests failed are once again not linked to your PR, i will merge it when i finished with pmx :)

toddwong:
Thanks,  @vmarchaud !

soyuka:
ping @wallet77

linonetwo:
Will this work in fork mode?

Cluster mode won't run ps1 file in windows. And fork mode still pop up cmd window.

Issue #3360, opened 2017-12-19
pm2 list and status show wrong user


pepez:
## What's going wrong?
pm2 list and status show current user instead of user running the application. In the example below the PID 47988 is running as user 501 (pm2-user).
```
[root@server ~]# ps aux | grep -i pm2
501      21566  0.0  0.7 936828 31252 ?        Ssl  Dec18   0:15 PM2 v2.7.2: God Daemon (/data/shop/.pm
root     48490  0.0  0.0 103300   800 pts/0    S+   10:04   0:00 grep -i pm2
[root@server ~]# ps aux | grep -i node
501      47988  0.4  1.4 1258148 59968 ?       Ssl  09:44   0:05 node /data/shop/shop-25/server-bundle.js      
root     48520  0.0  0.0 103296   792 pts/0    S+   10:05   0:00 grep -i node
[root@server ~]# pm2 list
┌───────────────┬────┬──────┬───────┬────────┬─────────┬────────┬─────┬───────────┬──────┬──────────┐
│ App name      │ id │ mode │ pid   │ status │ restart │ uptime │ cpu │ mem       │ user │ watching │
├───────────────┼────┼──────┼───────┼────────┼─────────┼────────┼─────┼───────────┼──────┼──────────┤
│ shop-25       │ 2  │ fork │ 47988 │ online │ 0       │ 20m    │ 0%  │ 58.5 MB   │ root │ disabled │
└───────────────┴────┴──────┴───────┴────────┴─────────┴────────┴─────┴───────────┴──────┴──────────┘
 Use `pm2 show <id|name>` to get more details about an app
[root@server ~]# su - pm2-user
[pm2-user@server ~]$ pm2 list
┌───────────────┬────┬──────┬───────┬────────┬─────────┬────────┬─────┬───────────┬────────────┬──────────┐
│ App name      │ id │ mode │ pid   │ status │ restart │ uptime │ cpu │ mem       │ user       │ watching │
├───────────────┼────┼──────┼───────┼────────┼─────────┼────────┼─────┼───────────┼────────────┼──────────┤
│ shop-25       │ 2  │ fork │ 47988 │ online │ 0       │ 20m    │ 0%  │ 58.5 MB   │ pm2-user   │ disabled │
└───────────────┴────┴──────┴───────┴────────┴─────────┴────────┴─────┴───────────┴────────────┴──────────┘
 Use `pm2 show <id|name>` to get more details about an app

```

```
$ pm2 report
===============================================================================
[...]

PR #3402, opened 2018-01-12
#3360 pm2 show wrong user, now display user who started the application

wallet77:
PM2 display user according to os.userInfo().username just before printing results.
At this moment the user is the one who launched PM2 Cli and not the one who started the application.

Move the code into executeApp function in God.js solves the problem.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3360
| License       | MIT

### DIFF
--- a/lib/God.js
+++ b/lib/God.js
@@ -16,8 +16,7 @@
 
 var semver        = require('semver');
 var cluster       = require('cluster');
-var os            = require('os');
-var numCPUs       = os.cpus() ? os.cpus().length : 1;
+var numCPUs       = require('os').cpus() ? require('os').cpus().length : 1;
 var path          = require('path');
 var EventEmitter2 = require('eventemitter2').EventEmitter2;
 var fs            = require('fs');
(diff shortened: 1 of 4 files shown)


lib/God.js
```diff
@@ -86,6 +87,21 @@ God.executeApp = function executeApp(env, cb) {
...
+      current_user = os.userInfo().username;
+    } catch (err) {
+      // For the case of unhandled error for uv_os_get_passwd
+      // https://github.com/Unitech/pm2/issues/3184
+      current_user = '';
+    } else {
+    current_user = process.env.USER || process.env.LNAME || process.env.USERNAME || process.env.SUDO_USER || process.env.C9_USER || process.env.LOGNAME;
+  }
+
+  env_copy['username'] = current_user;
```
vmarchaud: This part is executed on the daemon side so the current user will always use the one of the daemon.

Issue #3429, opened 2018-01-26
Feature Request: preserve old process when reload fails 


rokoroku:
Sometimes we **reload** the existing process with newly bundled (transpiled) code, but the new bundled code often break the application. 

In this case, **pm2** can use old process without changing its state as `errored`. 
Just like **graceful stop & reload**, if pm2 couldn't get `ready` signal before `listen_timeout` (or get `error`), it can stash the new process and preserve the old process without sending kill signal.

This is a feature like blue-green deployment.

After taking some look at the source, it could be achieved with a simple boolean flag like `reloadOnlyAfterReady`.

Could it be a PR welcomed feature?

wallet77:
Hi @rokoroku 

Yes it can definitely be a feature !
Do you want to take care of this ?

rokoroku:
I'll try.

`wait_ready` already sounds like a feature for this, however, it kills old process anyway. 
May I add a flag like `reload_after_ready` ? any suggestion?

pinkasey:
@Unitech 
Why was this Issue closed?
This feature is VERY MUCH IN NEED.

The behavior of "don't reload process X before process X-1 is done reloading _**SUCCESSFULLY**_" - is what we all want.
anyone who uses `pm2 reload` wants that.
No one wants their server to go completely down, during a bad deployment (deployment with a bug that prevents server from booting).

Won't anyone pick the glove from here?
After 7 years?

Issue #3407, opened 2018-01-16
PM2 reload in cluster mode kill current connections


quydo:
Hello team,

I have nodejs backend behind nginx proxy.

In nodejs backend I use PM2 for management, I started in cluster mode
ref: https://futurestud.io/tutorials/pm2-cluster-mode-and-zero-downtime-restarts

pm2 start index.js --name api -i 2 --watch  (1)

when I run command: pm2 reload api, there are some connections log 502 occur in nginx log, for example

OK 5.815 162.158.165.151 - - [16/Jan/2018:16:10:09 +0700] 502 "GET /api/v2/get_config?id=1235

I guess when I run command (1), pm2 will force x process and they can serve clients at this time, but It kill old processes (that serves some connections), instead of wait for they complete connections

I know nginx reload mechanism: nginx start new workers, all new connections will be served by these workers, and old workers will process old connections. After all old connections completed, it shutdown old nginx workers without downtime

Can I do this in PM2?

Thanks Team

wallet77:
Hi @quydo,

Did you try "gracefulReload" option ?

PR #3433, opened 2018-01-26
Add reload_after_ready flag to Reload (Draft)

rokoroku:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | ~yes~ no
| Fixed tickets | #3429 
| License       | MIT

This PR is a draft to achieve a feature I suggested in #3429 

I'm not sure how the flag should be like.. there are many options
1. `preserveOldWhenFail` - only resotre old process when new process crashes
2. `preserveOldWhenNoReady` - only resotre old process when new process does not send 'ready'
3. `stopWhenFail` - only stop the process when new process crashes: this is a BREAKING CHANGE that affects default behavior
... and more?

It's not a completed PR. I just want to ask the way from here.


lib/God/Reload.js
```diff
@@ -157,7 +157,22 @@ function hardReload(God, id, wait_msg, cb) {
...
+          if (err) return cb(err);
+
+          // Restore old worker
+          God.clusters_db[id] = old_worker;
+          delete God.clusters_db[t_key];
+
+          old_worker.pm2_env.pm_id = id;
+          old_worker.pm_id = id;
+
+          cb(null, proc);
```
rokoroku: Maybe I can throw an error here to tell reload process has been failed.

wallet77:
I think the best is between option 1 & 2.

Ideally we should find the best behavior to ensure 0 downtime.
Something like : 
1 - user try to reload or restart
2 - create a new worker (for each existing worker)
3 - if new worker fail or is not ready after a certain delay -> destroy it and keep old one
   - if new worker is ready, disable old one and redirect load to the new one + wait old worker to process old connections (see https://github.com/Unitech/pm2/issues/3407)

What do you think about that ?

rokoroku:
This PR is exactly for what you said.

However, I don't think it's required to copy previous files to another folder since **build process** is not the main feature handled by PM2. 

I want to know which flags should I add to this PR, and which additional behavior should added to this PR .

pinkasey:
@Unitech 
Why was this PR closed?
This feature is VERY MUCH IN NEED.

The behavior of "don't reload process X before process X-1 is done reloading _**SUCCESSFULLY**_" - is what we all want.
anyone who uses `pm2 reload` wants that.
No one wants their server to go completely down, during a bad deployment (deployment with a bug that prevents server from booting).

Won't anyone pick the glove from here?

Issue #1234, opened 2015-05-04
Is it possible to require process.send("online") before it goes online?


owenallenaz:
From the docs on graceful reload I see there is a way to nofity pm2 for non-http type processes that the process is online via a `process.send('online')`. Would it be possible to have a configuration setting that requires that to be sent in order for a process to be considered booted in all circumstances, and not just in graceful reloads.

This makes it very clear as developer where in the code you are notifying pm2 that your process is operational. In example, right now with graceful reload, pm2 assumes that my process is ready when it starts listening on something, but what if my process needs to listen to multiple things before it's "booted". In addition, when I first start it up I want to tell pm2 when I consider this boot a success. It's sort of an explicit over implicit type issue.

adrian-ludwig:
+1

jshkurti:
I get it. We have to think about it in order to avoid losing usability for the rest of the users. Also there is maybe a workaround : stop the 'listen' event to be fired automatically and fire it manually when your process is ready.

michaelBenin:
+1

Unitech:
```
$ pm2 start app.js --wait-ready --listen-timeout 15000
```

In your code, once your application is ready, do `process.send('ready')`

Available in pm2@next:

```
$ npm install pm2@next -g
$ pm2 update
```

PR #3511, opened 2018-03-02
feature: add inspect feature, on runtime

wallet77:
add inspector feature on process container.
Use pmx, not need to reload/restart app.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1234, #5678
| License       | MIT

### DIFF
--- a/lib/API/Extra.js
+++ b/lib/API/Extra.js
@@ -644,10 +644,15 @@ module.exports = function(CLI) {
     const that = this;
     if(semver.satisfies(process.versions.node, '>= 8.0.0')) {
       this.trigger(app_name, 'internal:inspect', function (err, res) {
-        if(res && res[0].data.return === '') {
-          Common.printOut(`Inspect disabled on ${app_name}`);
+
+        if(res && res[0]) {
+          if (res[0].data.return === '') {
+            Common.printOut(`Inspect disabled on ${app_name}`);
+          } else {
+            Common.printOut(`Inspect enabled on ${app_name} => go to chrome : chrome://inspect !!!`);
+          }
         } else {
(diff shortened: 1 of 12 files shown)


lib/ProcessContainerFork.js
```diff
@@ -1,17 +1,33 @@
...
+  const pmx = require('pmx');
+  pmx.init({
     transactions: (process.env.km_link === 'true' && (process.env.trace === 'true' || process.env.deep_monitoring === 'true')) || false,
     http: process.env.km_link === 'true' || false,
     v8: process.env.v8 === 'true' || process.env.deep_monitoring === 'true' || false,
     event_loop_dump: process.env.event_loop_inspector === 'true' || process.env.deep_monitoring === 'true' || false,
     deep_metrics: process.env.deep_monitoring === 'true' || false
   });
+
+  if(require('semver').satisfies(process.versions.node, '>= 8.0.0')) {
```
vmarchaud: We should also add the custom action in the `ProcessContainer` which is used in the cluster mode

lib/API/Extra.js
```diff
@@ -644,7 +644,7 @@ module.exports = function(CLI) {
     const that = this;
     if(semver.satisfies(process.versions.node, '>= 8.0.0')) {
       this.trigger(app_name, 'internal:inspect', function (err, res) {
-        if(res[0].data.return === '') {
+        if(res && res[0].data.return === '') {
```
vmarchaud: shouldn't we check that the `res[0]` is not undefined ?

wallet77: Done

Issue #1869, opened 2016-01-01
Support child process monitoring


jadbox:
I'm using the below standard lib to manually fork processes for my worker; however, those forks do not actually show up in the pm2 monit/list read out. This makes me think that pm2 is unable to detect that a script has manually forked. Is there a way to support showing these manual forks in list/monit?

https://nodejs.org/api/cluster.html

Unitech:
PM2 embeds the Node.js cluster logic for you, you do not need to do it manually. We may add the feature of monitoring child processes at some point but not for now,

Best

jadbox:
For my use-case, I require to do it manually. This is because the app is not a web server where ports can be routed between, but rather a worker that needs to fork at will depending on its job.

I'll be looking forward if pm2 supports monitoring when processes get forked internally.

soyuka:
For this to work, we'd need to access the child processes. I haven't found a way to automate this and child processes would have to notify PM2 they exists. 

I'd love this feature too and I'm looking for a way to implement it. 

Actually based on the above idea, api could look like this:

```
const pm2Child = require('pm2-child')

let child = fork('mychild.js')

//bind to pm2, with the fork object we can get back pid. args etc. (maybe streams)
pm2Child(child)
```

It should work the same way with cluster.

mattisx:
I'd love this too.

ochart2:
+1

MoLow:
+1

scippio:
+1

jackycute:
+1

soyuka:
We need to implement this IMO: https://github.com/soyuka/pm2-childprocess-test

Unitech:
I still think that using a command based on ps would allow to track the
number of child processes for any kind of app managed by pm2

On Sat, Dec 24, 2016 at 1:48 PM Antoine Bluchet <notifications@github.com>
wrote:

We need to implement this IMO:
https://github.com/soyuka/pm2-childprocess-test




—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
<https://github.com/Unitech/pm2/issues/1869#issuecomment-269083148>, or mute
the thread

Unitech:
Or directly via procfs

DevBrent:
[...]

PR #3554, opened 2018-03-19
Attempt to improve treekill

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |  na
| License       | MIT

Don't merge this for now.

Attempt to improve treekill performances (the `pstree` module could be used to get a child process monitoring on pm2 see this 2 years wanted feature https://github.com/Unitech/pm2/issues/1869). 
@simonepri has worked on the best way to get a tree of a given PID. He has found out that ps was faster then `pgrep`. Let's see how the test suite runs for now.

### DIFF
--- a/lib/TreeKill.js
+++ b/lib/TreeKill.js
@@ -1,41 +1,14 @@
 'use strict';
-
-// From https://raw.githubusercontent.com/pkrumins/node-tree-kill/master/index.js
-
-var childProcess = require('child_process');
-var spawn = childProcess.spawn;
-var exec = childProcess.exec;
+var pidtree = require('pidtree');
 
 module.exports = function (pid, signal, callback) {
-  var tree = {};
-  var pidsToProcess = {};
-  tree[pid] = [];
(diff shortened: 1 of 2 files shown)


simonepri:
Also here is not all about the performance but about correctness.

Actually the current implementation does not take care of nested subprocesses.
Both `pgrep -P` and `ps -o pid --ppid` **just** looks for the direct children of the given pid and does not return the whole subtree.

image

@Unitech what is the desidered behaviour?

simonepri:
Oh I didn't see that the command is called inside `buildProcessTree`.
So you recursively remove all the children.

This has the great advantage of not needing to parse the whole pid list as I do in `pidtree`.
I don't know if this actually improve performances. The overhead of spawning `ps` or `pgrep` a lot of time should kill any kind of advantage.

Issue #3739, opened 2018-06-26
[Question] Properly support nvm for Windows


JimiC:
## What's going wrong?
`pm2` does not properly support [`nvm` for Windows](https://github.com/coreybutler/nvm-windows)

Specifically:
- `pm2` does not detect the enviromnent variable used by `nvm` for Windows
- `pm2` does not detect the correct path to where `nvm` for Windows keeps the `node` versions

## How could we reproduce this issue?
- Install `nvm` for Windows
- Install a `node` version with `nvm`
- Try to run an app using `--interpreter node@x.x.x`

## Supporting information

```
$ pm2 report
```
```txt
[2018-06-26 14:55:39] PM2 log: ===============================================================================
[2018-06-26 14:55:39] PM2 log: --- New PM2 Daemon started ----------------------------------------------------
[2018-06-26 14:55:39] PM2 log: Time                 : Tue Jun 26 2018 14:55:39 GMT+0300 (GTB Daylight Time)
[2018-06-26 14:55:39] PM2 log: PM2 version          : 2.10.4
[2018-06-26 14:55:39] PM2 log: Node.js version      : 6.14.3
[2018-06-26 14:55:39] PM2 log: Current arch         : x64
[2018-06-26 14:55:39] PM2 log: PM2 home             : C:\Users\%user%\.pm2
[2018-06-26 14:55:39] PM2 log: PM2 PID file         : C:\Users\%user%\.pm2\pm2.pid
[2018-06-26 14:55:39] PM2 log: RPC socket file      : \\.\pipe\rpc.sock
[2018-06-26 14:55:39] PM2 log: BUS socket file      : \\.\pipe\pub.sock
[2018-06-26 14:55:39] PM2 log: Application log path : C:\Users\%user%\.pm2\logs
[2018-06-26 14:55:39] PM2 log: Process dump file    : C:\Users\%user%\.pm2\dump.pm2
[2018-06-26 14:55:39] PM2 log: Concurrent actions   : 2
[2018-06-26 14:55:39] PM2 log: SIGTERM timeout      : 1600
[2018-06-26 14:55:39] PM2 log: ===============================================================================
[2018-06-26 14:55:40] PM2 log: Starting execution sequence in -fork mode- for app name:app1 id:0
[2018-06-26 14:55:40] PM2 error: Error: spawn C:\Users\%user%\AppData\Roaming\nvm\versions\node\v6.14.3\bin\node ENOENT
    at exports._errnoException (util.js:1020:11)
[...]

PR #3748, opened 2018-06-28
Add support for nvm for Windows

JimiC:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | N/A
| Fixed tickets | #3739
| License       | MIT

These are the minimum required changes to support `nvm` for Windows.

### DIFF
--- a/lib/Common.js
+++ b/lib/Common.js
@@ -343,29 +343,42 @@ var resolveNodeInterpreter = function(app) {
     Common.printError(cst.PREFIX_MSG_WARNING + chalk.bold.yellow('Choosing the Node.js version in cluster mode is not supported'));
     return false;
   }
-  if (!process.env.NVM_DIR) {
+
+  var nvm_path = cst.IS_WINDOWS ? process.env.NVM_HOME : process.env.NVM_DIR;
+  if (!nvm_path) {
     Common.printError(cst.PREFIX_MSG_ERR + chalk.red('NVM is not available in PATH'));
     Common.printError(cst.PREFIX_MSG_ERR + chalk.red('Fallback to node in PATH'));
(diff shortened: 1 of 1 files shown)


JimiC:
I'd like to explain lines 380-381 a bit as `nvm` for Windows has a quirk. 
When it first downloads `node.exe` for a version, it renames it to `node64.exe` for `x64` platforms and `node32.exe` for `x32` ones, in order to support both arch side by side.
Upon activation via `nvm use` it renames `nodeX.exe` to `node.exe` and remains so if no other arch version for the same `node` version is downloaded.

Now, I wanted to refrain from making the codebase complex by introducing filename checks, so I kept it at a bare minimum which leads to a quirky flow.

Explaining, `pm2` will always look for a `node.exe` filename when trying to resolve the specified interpreter. 
- If the user has already activated the `node` version, `pm2` will see that and proceed as normal.
- If the user has not downloaded the `node` version specified, `pm2` will trigger an `nvm install`. As mentioned above `nvm` for Windows has that renaming quirk so after the download is finished `pm2` will look for a `node64.exe` or `node32.exe` instead of `node.exe`, as the user has not yet activated that `node` version.
- If the user has previously downloaded the `node` version specified but has not activated it yet, the filename will still be `nodeX.exe`. `pm2` will look for a `node.exe` filename but will not find it and will try to download that version again. Now, `nvm` for Windows will see that that `node` version is already there and prompt that the version already exists and skips download. This is when lines 380-381 come to the rescue as they tell `pm2` to look for a `nodeX.exe` filename instead, which is true and `pm2` then proceeds as normal.

If you think that this flow will cause maintainability issues, and you prefer a filename check function to be added instead, please let me know, so to update the PR.

JimiC:
Unfortunately, no tests have been added, as you don't have a `Windows` testing environment, yet.

PR #3784, opened 2018-07-11
feature: pm2 plus cli

Unitech:


lib/API/Modules/Modularizer.js
```diff
@@ -68,19 +68,19 @@ Modularizer.install = function (CLI, moduleName, opts, cb) {
       return cb(Common.retErr(e));
     }
 
-    Modularizer.installMultipleModules(config.dependencies, cb);
+    Modularizer.installMultipleInternalModules(config.dependencies, cb);
```
wallet77: This method can be used to install multiple modules from "dependencies" property in ecossytem file.
So it can install modules which are not "internals".

Unitech: I specified a name for it because it only install internal module and not independent module (like pm2-server-monit, pm2-logrotate...)

Issue #3786, opened 2018-07-12
pm2 trigger command hangs


pankleks:
I sending custom command to pm2 process with `pm2 trigger myapp test`

After issuing command pm2 answers:

```
[myapp:4]="my answer"
1 processes have received command test
```

Then it hangs.

This does not happen everytime, in most cases it works - but I can't find any pattern.

System: Ubuntu 16.04, 18.04
PM2: 2.6, 3.0.0

In my app I respond according to doc:

```ts
        Pmx.action("test", (reply) => {
            reply(`my answer`);
        });
```

Unitech:
Are you using https://www.npmjs.com/package/pmx or https://www.npmjs.com/package/@pm2/io?
Which pmx version?

Thanks

pankleks:
Thanks for reply. I'm using Pmx

pankleks:
Not sure how to check pmx version, I'm adding package `pm2` in version 2.10.4

Unitech:
action.js: 

```
var pmx = require('pmx')

pmx.action("test", (reply) => {
  reply(`my answer`);
});
```

```
$ pm2 start action.js 
$ pm2 trigger action test
c1 processes have received command test
[action:0]="my answer"
```

PMX version 1.6.7, PM2 3.0

I did the command 20 times it worked every time.

How did you started your application?

pankleks:
it's installed as pm2 module with `pm2 install my-app`

wallet77:
It seems we have this behavior if the called action doesn't exist.
I will try to fix it.

Anyway if the command hangs so your action is not detected and probably not correctly initialized.

PR #3874, opened 2018-08-21
fix: #3786 fix issue when triggering an action that does not exist

wallet77:
If action doesn't exist we should display an error, even if we trigger action by application's name.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3786
| License       | MIT

### DIFF
--- a/lib/God/ActionMethods.js
+++ b/lib/God/ActionMethods.js
@@ -801,17 +801,11 @@ module.exports = function(God) {
 
         var proc_env = God.clusters_db[id].pm2_env;
 
-        var action_exist = false;
-        proc_env.axm_actions.forEach(function(action) {
-          console.log(action, cmd)
-          if (action.action_name === cmd.msg) {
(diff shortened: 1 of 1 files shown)


lib/God/ActionMethods.js
```diff
@@ -801,6 +801,22 @@ module.exports = function(God) {
 
         var proc_env = God.clusters_db[id].pm2_env;
 
+        var action_exist = false;
+        proc_env.axm_actions.forEach(function(action) {
```
vmarchaud: Could be refactored to a simpler code : 

```javascript
const isActionAvailable = proc_env.axm_actions.find(action => action.action_name === cmd.msg) !== undefined
```

There is `console.log` call also

wallet77: done

PR #3907, opened 2018-09-12
plus: fix oauth url + add terms & conditions in register

vmarchaud:


vmarchaud:
Test are failing but not related to the changes, could you checkout why @wallet77 ?

PR #3912, opened 2018-09-14
Cpu limit

deanrather:
Adds support for an optional `cpu_limit` option on applications defined in the environment.config.js file.

eg:

```js
// environment.config.json
module.exports = {
  apps: [{
      script: './heavy-service.sh',
      args: ['--port=1111', '--name=service1'],
      cpu_limit: 1
    },
    {
      script: './heavy-service.sh',
      args: ['--port=2222', '--name=service2'],
      cpu_limit: 1
    },
    {
      script: './heavy-service.sh',
      args: ['--port=3333', '--name=service3'],
      cpu_limit: 1
    },
    {
      script: './heavy-service.sh',
      args: ['--port=4444', '--name=service4'],
      cpu_limit: 10
    }, {
      script: './heavy-service.sh',
      args: ['--port=5555', '--name=service5'],
      cpu_limit: 10
    }
  ]
};
```

```sh
#!/bin/bash
# heavy-service.sh

example_arg1=$1
example_arg2=$2

# this will use up all available CPU

while true; do
  echo "heavy-service running with example_arg1: $example_arg1, example_arg2: $example_arg2"
  # sleep 0.000001
done
```

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT


Unitech:
good PR idea
We've noticed that since Node 10.10 there is os.{get,set}Priority https://github.com/nodejs/node/pull/22407
It would be better to leverage this API instead

Issue #2487, opened 2016-11-01
Cron on offline processes


Unitech:
Currently the cron_restart option only works for online processes ([[1](https://github.com/Unitech/pm2/blob/master/lib/ProcessContainer.js#L92)][[2](https://github.com/Unitech/pm2/blob/master/lib/ProcessContainerFork.js#L19])])

That logic should be moved to the Worker so It can restart offline applications (useful when used with autorestart to false)

Almya:
+1

khromov:
Also looking for this feature. Has there been any movement on it?

tpalfr:
Any news on this feature ?

arthursp:
Any new

mattb290:
Does this also cause memory to not be released? I have a node program which is being run every hour using pm2 cron, and it completes after about 20 seconds. However for the rest of the hour, pm2 list shows it as taking up ~50mb of memory.

soyuka:
My workaround while this isn't fixed:

1. Create a `bin/job` that will execute your script for example:

```javascript
#!/usr/bin/env node
const { exec } = require('child_process')

exec(process.argv.slice(2).join(' '), function (error, stderr, stdout) {
  if (error) {
    console.error(`exec error: ${error}`);
    return;
  }

  console.log(`stdout: ${stdout}`);
  console.log(`stderr: ${stderr}`);
})

// fake long running process
setInterval(() => {}, 1e6)
```

Then this is the process declaration:

```json
{
    "name": "cron_compute_midnight",
    "script": "./bin/job",
    "exec_mode": "fork",
    "args": "curl -s http://localhost/api/batch/compute",
    "cron_restart": "0 0 * * *"
  }
```

Obviously, if this issue were to be fixed the following would work:

```json
{
    "name": "cron_compute_midnight",
    "script": "curl",
    "exec_mode": "fork",
    "args": "-s http://localhost/api/batch/compute",
    "cron_restart": "0 0 * * *",
    "autorestart": false
  }
```

PR #3944, opened 2018-10-06
Allow to use cron with autorestart=false scripts

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2487  
| License       | MIT
| Doc PR        | na

To test:
```
./bin/pm2 start -c '* * * * *' --no-autorestart --no-daemon test/fixtures/cron_no_autorestart.js
```

Wait at least 1 min

### DIFF
--- a/lib/Worker.js
+++ b/lib/Worker.js
@@ -13,7 +13,7 @@ var cronJob = require('cron').CronJob
 module.exports = function(God) {
   var timer = null;
 
-  God.CronJobs = {};
+  God.CronJobs = new Map();
   God.Worker = {};
   God.Worker.is_running = false;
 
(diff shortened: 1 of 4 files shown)


Unitech:
Wise fix
It allows to use cron on stopped apps + on any kind of process besides node.js

soyuka:
> Wise fix
> It allows to use cron on stopped apps + on any kind of process besides node.js

And fixed a 3 years old issue :fireworks:

lib/Worker.js
```diff
@@ -20,6 +22,34 @@ module.exports = function(God) {
...
+    if (!proc ||
+        !proc.pm2_env ||
+        proc.pm2_env.pm_id === undefined ||
+        !proc.pm2_env.cron_restart ||
+        God.CronJobs[proc.pm2_env.pm_id] !== undefined)
+      return;
+
+    console.log('[PM2][WORKER] Registering a cron job on:', proc.pm2_env.pm_id);
+
+    God.CronJobs[proc.pm2_env.pm_id] = new cronJob({
```
vmarchaud: I believe a Map should be better here

soyuka: Are maps supported by node 4?

vmarchaud: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Browser_compatibility
Since node 0.12

soyuka:
> Programmatic tests are missing, checkout https://github.com/Unitech/pm2/blob/master/test/programmatic/exp_backoff_restart_delay.mocha.js

I can but the test will have to wait at least 1 whole minute to check that the cron did his job.

PR #4018, opened 2018-11-11
pm2 should ignore -s, --silent, -v after '--'

AlexanBor:
pm2 should ignore -s --silent -v if they are after '--' (variadic args for start and restart commands)
Example:
pm2 start node-red --node-args="--max-old-space-size=128" -- -v
Currently just prints versions and exits.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT


Unitech:
good catch :)

PR #4099, opened 2019-01-06
feat: re-implement pm2 plus cli

vmarchaud:


lib/Client.js
```diff
@@ -239,13 +239,9 @@ Client.prototype.launchDaemon = function(opts, cb) {
   if (!process.env.PM2_DISCRETE_MODE)
```
Unitech: Could you remove edits on this Client.js file

Issue #4150, opened 2019-02-04
systemd.tpl for Ubuntu startup is missing restart lines


sirisian:
This might be related to #3866, but I noticed that when OOM killer in Ubuntu kills pm2 it doesn't restart. There are comments online that startup should restart, so I looked at the .service file in ```/etc/systemd/system/pm2-<user>.service``` and it's missing the lines:

```
[Service]
Restart=always
RestartSec=30
```

The template for where this should be is here:

https://github.com/Unitech/pm2/blob/master/lib/templates/init-scripts/systemd.tpl

Am I misunderstanding the startup script? It seems like pm2 should restart by default if OOM killer kills it randomly. Or is this the intended behavior that it stops permanently until reboot?

PR #3956, opened 2018-10-12
Auto restart god

deanrather:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        | https://github.com/pm2-hive/pm2-hive.github.io/pulls

updates the systemd template files to auto-restart the god daemon should it somehow crash.

you can test with `sudo kill -9 <pid of god daemon>`

pm2 doesn't restart itself! even after `pm2 startup`.

this fixes that.

(also, I accidentally included a commit where I reduced the .deb packages node dep from 6.12.3 down to 6.3.0 -- that's not required for this MR, but it does help me as I've got a lot of servers w/ 6.3.0 installed...)

PR #4157, opened 2019-02-09
Added Restart and RestartSec into the systemd script

sirisian:
When Ubuntu's out of memory killer kills PM2 it never restarts. Adding these lines should restart it after 30 seconds.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #4150 
| License       | MIT


Unitech:
this is a tricky one, because sometime systemd does not detect that PM2 is already launched and it might respawn another PM2 :/
Going to merge it but I have to make to tests

Unitech:
check this, we've already tried to implement this but it was reverted: https://github.com/Unitech/pm2/commit/1f855da544be9625ec0ff38b2dfbd0d8a3df8247#diff-46f23e2b9f1602cc89e4e73cb40132d7

Unitech:
`Restart=on-failure` works much better!
https://github.com/Unitech/pm2/pull/3956
thanks @deanrather

Issue #3732, opened 2018-06-22
pm2 cannot delete launchd script on mac


micuat:
## What's going wrong?
`unstartup` fails with undefined error. launchd keep opens pm2 after boot (moved from https://github.com/Unitech/pm2/issues/1349 as this is a new issue).

## How could we reproduce this issue?
(from https://github.com/Unitech/pm2/issues/1349#issuecomment-397797391)
```
➜  ~ pm2 unstartup
[PM2] Init System found: launchd
[PM2] To setup the Startup Script, copy/paste the following command:
sudo env PATH=$PATH:/usr/local/bin /usr/local/lib/node_modules/pm2/bin/pm2 unstartup launchd -u theara --hp /Users/theara
➜  ~ sudo env PATH=$PATH:/usr/local/bin /usr/local/lib/node_modules/pm2/bin/pm2 unstartup launchd -u theara --hp /Users/theara
Password:
[PM2] Init System found: launchd


undefinedReturn code : 3
```

## Supporting information

I found that deleting `~/Library/LaunchAgents/pm2.*` will stop launching the daemon. The problem is that the plist is named `pm2.username.plist` but the script is trying to delete the file `com.pm2.username.plist`
https://github.com/Unitech/pm2/blob/f8a6a634623808625c9a0861553307268878944a/lib/API/Startup.js#L140

micuat:
seems like the PR already fixed it? closing...

Issue #1349, opened 2015-06-13
Document how to remove startup script


dandv:
https://github.com/Unitech/pm2#startup-script-generation talks about generating startup script. What about removing the scripts for an app that's no longer needed?

ilaif:
:+1: 
I tested it on my mac to make sure it works. Now I want to remove it and I have no idea how.

catdad:
Agreed that this is absolutely needed. I want to use PM2 to run a service from an installer, and need to remove the script on uninstall. Documentation would be great, but a built-in command is even better.

khaled-hossain-code:
i do need to remove my startup script in debian. how to do it with pm2?

Unitech:
On PM2 > 2.2.x:

```
$ pm2 unstartup
```

Netoperz:
Not useful,
Problem presists.

When you install pm2 using apt-get install

And than uninstall using apt-get uninstall 

the errors are shown on boot screen, because the scripts presists

and there is no pm2 on system already to run

pm2 unstartup

.....

chrisli30:
Anybody got it working?
First I ran this.
pm2 unstartup
And got,
[PM2] Init System found: launchd
[PM2] You have to run this command as root. Execute the following command:
sudo env PATH=$PATH:/usr/local/bin /usr/local/lib/node_modules/pm2/bin/pm2 unstartup launchd -u username --hp /Users/username

Then I run the command above and got undefined return code: 3. 
[PM2] Init System found: launchd
undefinedReturn code : 3

Any help would be appreciated.

kuzmicheff:
@chrisli30 I have started using pm2 very recently, and I have run into the same message trying to remove the startup script on my local Mac OS 10.12. My problem is that as soon as I ran the startup script and restarted the laptop, the memory usage stopped working. It seems that it fails to read the memory stats, but I haven't figured out the source of the problem yet. I could not find any relevant information online so I figured I would look at my config and system to try to pinpoint it more precisely before submitting a new issue on GitHub. It seems you are running pm2 on Mac OS as well. Have you observed any similar symptoms upon installing the startup script? Thanks!

chrisli30:
Sorry, @kuzmicheff no I haven't dived much into memory usage feature, but mine seems to work fine on KeyMetrics which uses PM2. One thing to look at though is the same problem happen on another machine running Linux?

kuzmicheff:
[...]

PR #4197, opened 2019-03-12
fix pm2 unstartup fialed in macos

harlentan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3732 , #1349 
| License       | MIT

The `pm2 unstartup` will always failed in macos. Because the pm2.*.plist will never been deleted.

```
commands = [
        'launchctl remove ' + launchd_service_name + ' || true',
        'rm ' + destination
      ];
```
`launchctl remove` always return no-zero value, so `rm` in the next will never been executed.


harlentan:
How to fix the CI build failed? I think there is nothing to do with my patch.

PR #3723, opened 2018-06-18
Add directory support for --ignore-watch, fixed --ignore-watch, implemented test and implemented flag --watch-delay 

livankrekh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | #1234, #5678
| License       | MIT
| Doc PR        | https://github.com/pm2-hive/pm2-hive.github.io/pulls

Issue #2097, opened 2016-04-16
ignore-watch not work


extensionsapp:
Hello.

My code is a change config file, but if PM2 watch changed in config.new.js and RESTARTED, and script fail.

```
fs.writeFile(path.join(path.dirname(__dirname), 'config', 'config.new.js'), 'module.exports = ' + data + ';', function(err) {
        if (err) {
            console.log(err);
            res.status(404).send('Error');
        }
        else {
            fs.rename(path.join(path.dirname(__dirname), 'config', 'config.js'), path.join(path.dirname(__dirname), 'config', 'config.old.js'), function (err) {
                if (err) {
                    console.log(err);
                    res.status(404).send('Error rename config.old.js');
                }
                fs.rename(path.join(path.dirname(__dirname), 'config', 'config.new.js'), path.join(path.dirname(__dirname), 'config', 'config.js'), function (err) {
                    if (err) {
                        console.log(err);
                        res.status(404).send('Error rename config.new.js');
                    }
                    res.status(200).send('Save');
                });
            });
        }
    });
```

I run app:

`pm2 start app.js --watch --name="example.com" --ignore-watch "node_modules config/config.new.js config/config.old.js" --no-vizion`
or
`pm2 start app.js --watch --name="example.com" --ignore-watch "node_modules config/config\.new\.js config/config\.old\.js" --no-vizion`
or
`pm2 start app.js --watch --name="example.com" --ignore-watch "config" --no-vizion`
or
`pm2 start app.js --watch --name="example.com" --ignore-watch="config" --no-vizion`

Not work! PM2 restarted app if changed config file.

soyuka:
To help debugging :

```
DEBUG="pm2:*" pm2 start --no-daemon --ignore-watch "config"
```

https://github.com/Unitech/PM2/issues/918

This option works and is tested.

extensionsapp:
How to ignore .txt files?

`
DEBUG="pm2:*" pm2 start app.js --no-daemon --watch --ignore-watch ".git \.txt" --name="domain.com"
`
or
`
DEBUG="pm2:*" pm2 start app.js --no-daemon --watch --ignore-watch ".git .txt" --name="domain.com"
`
or
`
DEBUG="pm2:*" pm2 start app.js --no-daemon --watch --ignore-watch ".git ^.*\.txt$" --name="domain.com"
`
or
`
[...]

PR #4224, opened 2019-03-29
Fix --ignore-watch to accept globs

vaskevich:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2097
| License       | MIT

It looks like #3723 broke `--ignore-watch`, as it now only really accepts a single directory and always overwrites the `ignore_watch` configuration with any files it finds recursively in that directory. I'm not sure why that was added since writing space-separated globs like `--ignore-watch "node_modules **/*_test.js"` seems to work pretty well.


Unitech:
thanks for your contribution
@soyuka wdyt?

Unitech:
works great
kudos to @vaskevich

Issue #3884, opened 2018-08-29
ANSI escape code not working after updated to PM2 v3


nammaianh:
I'm using PM2 v3.0.4 on OS X. I run `pm list --watch` but some ANSI code is not processed.

![image](https://user-images.githubusercontent.com/9427046/44765038-0b153100-ab7d-11e8-947a-f05d7c4d16a8.png)

Unitech:
Available in PM2 4.1.0

```bash
npm install pm2@latest -g
pm2 update
```

PR #4270, opened 2019-05-07
fix: ANSI escape codes with --watch #3884

taneltm:
Running `pm2 list --watch` doesn't update the status table properly.

The escape code should be `\033`, but that doesn't work in strict mode, probably because of that it was changed to `\\033` which isn't valid either.

Tests failed
```
[~] Starting test ./test/e2e/misc/nvm-node-version.sh
######## ✘ Errors in setting interpreters
```
But that doesn't seem to be related to this change.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | #3884
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/lib/API.js
+++ b/lib/API.js
@@ -589,8 +589,8 @@ class API {
       if (opts && opts.rawArgs && opts.rawArgs.indexOf('--watch') > -1) {
         var moment = require('moment');
         function show() {
-          process.stdout.write('\\033[2J');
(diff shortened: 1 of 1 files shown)


Unitech:
awesome thanks

PR #4288, opened 2019-05-22
add 'max' for instances option

sunecosuri:
ref: http://pm2.keymetrics.io/docs/usage/cluster-mode/#usage

Added because the type of instance option needs to include max

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        |

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -284,7 +284,7 @@ interface Pm2Env {
   /**
    * The number of running instances.
    */
-  instances?: number;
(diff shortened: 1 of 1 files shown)


Unitech:
Available in PM2 4.1.0

```bash
npm install pm2@latest -g
pm2 update
```

PR #4297, opened 2019-05-27
Support Solaris SMF as init system

licson0729:
Currently, when running the command `pm2 startup` on Oracle Solaris or illumos-based systems like SmartOS, OpenIndiana, etc. will throw up an `Init system not found` error.

This PR adds support for using the Solaris SMF system for starting PM2 which is available on Solaris 10 or above that replaces `/etc/init.d` scripts on such platform.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A


Unitech:
thanks

PR #4362, opened 2019-07-21
exec-mode as argument

LuKks:
`pm2 start app.js -i 1` starts as fork, it should be cluster but can be understandable due config files, etc.
After you can `pm2 scale app 2`, it still as fork mode and can there errors like EADDRINUSE.

The workaround can be:
`pm2 start app.js -i 2 && pm2 scale app 1`
Will start two instances in first place so it's not fine.

With this changes you can:
`pm2 start app.js -i 1 --exec_mode cluster`

Or simply due the default instance value:
`pm2 start app.js --exec_mode cluster`

Why?
Can't have no downtime with fork mode (and native cluster) due sending two signals forces the exit.
Also there is no way to start as cluster with 1 instance, at least from command line.

I'm worry about if it's need more changes but seems to works fine.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT


Unitech:
https://github.com/Unitech/pm2/commit/b122dca76488b23e57b3e6950b1f8a95715ad675

Unitech:
Available in PM2 4.1.0

```bash
npm install pm2@latest -g
pm2 update
```

PR #4380, opened 2019-08-02
Optionally expand variables in command arguments

crishoj:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

This PR adds the boolean option `expand_args` to the app configuration. When enabled in fork mode, shell-style variables in command arguments are expanded using the current environment.

Example, used in conjunction with `increment_var`:

```json5
{
  script: "bin/server",
  instances: 4,
  exec_mode: "fork_mode",
  increment_var: 'PORT',
  env: {
    PORT: 8000
  },
  args: '--port $PORT',
  expand_args: true
}
```

Each spawned process receives a unique port number on the command line:

```sh
bin/server --port 8000
bin/server --port 8001
bin/server --port 8002
bin/server --port 8003
```


crishoj:
There may be a problem in this feature: 

While the variable expansion work when a process is first started, it seems to not work on restarts (with `pm2 restart N` or `pm2 stop N` followed by `pm2 start N`). 

I may need to ask your help debugging this issue. Is `pm2_env.env` no longer available at the point of restarts?

crishoj:
When a process is restarted, `pm2_env.expand_args` in `God.forkMode` is suddenly false. 

Other entries, such as `pm2_env.increment_var`, are retained, so I'm rather puzzled how `expand_args` gets cleared somewhere in between `start` and `restart`.

Unitech:
Try to add the key `increment_var` here: https://github.com/Unitech/pm2/blob/master/lib/Common.js#L513

crishoj:
Thanks. No luck — `pm2_env.expand_args` keeps getting reset to `false` after the first `restart`.

PR #4439, opened 2019-10-04
Fix watch argument when used in another program (ie -- --watch)

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | na
| License       | MIT
| Doc PR        | na

Sometimes when running a command inside pm2 you could do something like

```
pm2 start node -- npm run builder -- --watch
```

This would mean that the `builder` script will receive the `--watch` argument by convention. 
When PM2 received these argument, he would start the process and watch for changes but if we wanted that we should do:

```
pm2 start --watch node -- npm run builder -- --watch
```

This PR fixes the wrong behavior using `minimist` to parse arguments. Note that this was already a peer dependency of the project (installation time should not vary).

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -4,16 +4,17 @@
  * can be found in the LICENSE file.
  */
 
-var debug          = require('debug')('pm2:client');
-var Common         = require('./Common.js');
-var KMDaemon       = require('@pm2/agent/src/InteractorClient');
-var rpc            = require('pm2-axon-rpc');
-var forEach        = require('async/forEach');
-var axon           = require('pm2-axon');
-var util           = require('util');
-var fs             = require('fs');
-var path           = require('path');
-var pkg            = require('../package.json');
+var debug            = require('debug')('pm2:client');
+var Common           = require('./Common.js');
+var KMDaemon         = require('@pm2/agent/src/InteractorClient');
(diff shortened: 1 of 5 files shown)


Unitech:
The issue is that a new dependency has been added "minimist". Would you mind finding an alternative? Thank you

soyuka:
> Note that this was already a peer dependency of the project (installation time should not vary).

lib/Client.js
```diff
@@ -509,9 +510,12 @@ Client.prototype.executeRemote = function executeRemote(method, app_conf, fn) {
...
   }
-  else if (method.indexOf('restartProcessId') !== -1 && process.argv.indexOf('--watch') > -1) {
-    delete app_conf.env.current_conf.watch;
-    this.toggleWatch(method, app_conf);
+  else if (method.indexOf('restartProcessId') !== -1) {
+    if (app_conf.env.current_conf.watch === true || hasWatchArgument(process.argv)) {
+      this.toggleWatch(method, app_conf);
+    } else {
+      delete app_conf.env.current_conf.watch;
+    }
```
soyuka: This is wrong I need to fix it

Issue #3691, opened 2018-05-30
sendSignal SIGINT is terminating the process without graceful (even with --kill-timeout 3000)


apoguy:
## What's going wrong?
Using pm2 sendSignal SIGINT <process name> is terminating the process, without running the code in the app that listens for the SIGINT, and exiting with code 1.

## How could we reproduce this issue?
My node code for reading the SIGINT:
```
process.on("SIGINT", function () {
    //graceful shutdown
    server.end().then(() => {
        process.exit();
    }).catch((err) => {
        console.error(err);
    });

});
```
Im running pm2 like this:

> pm2 start server.js --name AppName --silent --kill-timeout 3000 --no-autorestart --disable-logs

When sending `pm2 sendSignal SIGINT AppName` the process ends with exit code 1, while if I run the app using node server.js and use CTRL+C, the app does catch the SIGINT and follows the correct behavior.

## Supporting information
```
PM2        | [2018-05-30 13:53:11] PM2 log: Starting execution sequence in -fork mode- for app name:AppName id:0
PM2        | [2018-05-30 13:53:11] PM2 log: App name:AppName id:0 online
PM2        | [2018-05-30 13:53:58] PM2 log: Stopping app:AppName id:0
PM2        | [2018-05-30 13:53:58] PM2 log: App [AppName] with id [0] and pid [15176], exited with code [1] via signal [SIGINT]
PM2        | [2018-05-30 13:53:58] PM2 log: pid=15176 msg=process killed
```

aeisenberg:
I think we are having a similar issue. When our apps are in cluster mode, graceful shutdown seems to forcefully end a process before the kill-timeout is complete.

Have you determined the cause or found a solution here?

aeisenberg:
Actually, on a deeper inspection. It looks like the graceful shutdown was actually working, but our log messages were not flushed to the log file before the final shutdown occurred. Once we added some code to wait for logging to be complete before shutting down everything appears to be happening normally.

It's just odd that the behavior is different in fork mode vs cluster mode.

Issue #3651, opened 2018-05-07
CTRL-C graceful exit vs PM2 stop under Windows


xorbis:
## What's going wrong?

When I press CTRL-C while running my NodeJS under Windows 10 64 bits, my shutdown sequence executes properly. But on the other hand, using pm2 stop doesn't call the shutdown sequence at all.

## How could we reproduce this issue?

Run a NodeJS program with process.on("SIGINT") under Windows.
I am also monitoring for SIGBREAK, SIGUSR2 and SIGTERM.

```
		// Constants
		this.signals = [
			"SIGBREAK",
			"SIGINT",
			"SIGUSR2",
			"SIGTERM"
		];
```

```
		// IPC signals
		this.signals.forEach((signal) => {
			this.debug("Adding hook on signal " + signal);

			process.once(signal, () => {
				this.debug(signal + " received");
				this.exit(signal);
			});
		});
```

I also have this snippet in the code:

```
		// NodeJS events
		process.on("message", (msg) => {
			this.dump(msg, "MSG");

			switch (msg) {
				case "shutdown":
					this.exit(msg);
					break;
			}
		});
```

But it is never called.

Also this one found on Stackoverflow:

```
		// Windows process signals IPC fallback code
		if (process.platform === "win32") {
			this.log("Running on the Windows 32 platform");

			let rl = require("readline").createInterface({
			  input: process.stdin,
			  output: process.stdout
			});

			this.signals.forEach((signal) => {
				this.debug("Forwarding win32 signal " + signal + " to NodeJS");

				rl.on(signal, () => {
					this.debug("Received win32 signal " + signal);
			  		process.emit(signal);
				});
			});
		}
```

## Supporting information

```
===============================================================================
--- PM2 REPORT (Mon May 07 2018 10:30:34 GMT-0400 (Eastern Summer Time)) ------
===============================================================================
--- Daemon -------------------------------------------------
pm2d version         : 2.10.2
node version         : 8.9.1
node path            : C:/Program Files/nodejs/node
argv                 : C:\Program Files\nodejs\node.exe,C:\Users\masse\AppData\Roaming\npm\node_modules\pm2\lib\Daemon.js
argv0                : node
user                 : undefined
uid                  : N/A
gid                  : N/A
uptime               : 15min
===============================================================================
--- CLI ----------------------------------------------------
local pm2            : 2.10.2
node version         : 8.9.1
node path            : undefined
[...]

Issue #3555, opened 2018-03-19
Windows - process cannot catch PM2 `SIGINT`


braytak:
## What's going wrong?
My `Windows` process is seemingly not getting `SIGINT` on `stop` command, since my `process.on('SIGNINT'...)` handler never gets triggered - so can't do gracefull exit! 
The process correctly handles `SIGINT` when run from a terminal window by doing `Ctlr-C`.

Furthermore, the `pm2` log head below shows that the process exited with code [1], while my signal handler exits with code [0] on  `SIGINT`. 

## How could we reproduce this issue?
Run the following with `pm2 start --no-autorestart   sigint.js` 

```
*File: sigint.js*

process.on('SIGINT', ()=>{
    clearInterval(interval);
    console.log('GOT SIGINT');
    setTimeout(()=>{process.exit(0)}, 1000)
});

let interval = setInterval(()=>{
    console.log('Waiting for SIGINT')
}, 5000)
```
You never see the `GOT SIGINT` in the process log... only a series of `Waiting for SIGINT` 

## Supporting information
`pm2` version : 2.10.1   
`node` version: 8.9.1
`Windows 10`

`pm2` log head:
```
[2018-03-19 14:29:38] PM2 log: Starting execution sequence in -fork mode- for app name:sigint id:0
[2018-03-19 14:29:38] PM2 log: App name:sigint id:0 online
[2018-03-19 14:29:46] PM2 log: Stopping app:sigint id:0
[2018-03-19 14:29:46] PM2 log: App [sigint] with id [0] and pid [14224], exited with code [1] via signal [SIGINT]
[2018-03-19 14:29:46] PM2 log: pid=14224 msg=process killed
```

fanfan6ji581:
I m having same problem

wallet77:
Hi there,

As nodejs doc says, Windows can't handle signals as Linux does : 

> Note: Windows does not support sending signals, but Node.js offers some emulation with process.kill(), and ChildProcess.kill(). Sending signal 0 can be used to test for the existence of a process. Sending SIGINT, SIGTERM, and SIGKILL cause the unconditional termination of the target process.

braytak:
@wallet77 I have to admit knowing that when I posted - my objective was to see if any of the bright minds out there had a workaround...

Since then I found this related piece of information in the `pm2` online doc; [Windows graceful stop](pm2.keymetrics.io/docs/usage/signals-clean-restart/).

[...]

PR #4474, opened 2019-10-24
Adding optional switch to enable graceful shutdown on Windows

aleksk:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #3651, #3691, #3555
| License       | MIT

---

*Problem*:
There is no official and easy way to implement the graceful shutdown on Windows.

For graceful shutdown PM2 pretty much relies on the `process.kill()` and applications being able to ignore `SIGINT` to delay shutdown, but it doesn't work on Windows.
https://pm2.keymetrics.io/docs/usage/signals-clean-restart/

There are variables you can set to alter the PM2 kill behavior: `PM2_KILL_SIGNAL` (`SIGINT` by default), `PM2_KILL_TIMEOUT` (`1600` by default), the `--treekill` parameter (`true` by default), and kill retry time.

---

There are a couple of issues with PM2's graceful shutdown on Windows:
* The `shutdown` message is only sent if the `softReload` method is used: https://github.com/Unitech/pm2/blob/bf1c605fd40a8b4ec5f76aca95debe94bdd7a357/lib/God/Reload.js#L77 I.e. it works only if you call `pm2 startOrGracefulReload <json>`, and if the application is automatically restarted by the PM2. **But** it does not work for kill/stop/restart/reload **and** for scale down operations.
* People are often confused about `process.kill()` behavior on Windows due to unclear documentation (https://nodejs.org/api/process.html#process_process_kill_pid_signal) - Windows doesn't support signals; instead, nodejs emulates `process.kill()` behavior for 3 signals and ignores everything else: `SIGINT`/`SIGTERM` - terminates app; `0` - checks whether the process exists. What is often confusing is that if you run nodejs app in cmd and use `Ctrl-C`, it _does_ receive `SIGINT`, - but it's emulated.

---

*Attempted solutions*:
1. Existing bugreports. There a bunch of posts on stackoverflow/github about emulating `SIGINT`/`SIGBREAK` by handling `STDIN` closure: https://stackoverflow.com/questions/10021373/what-is-the-windows-equivalent-of-process-onsigint-in-node-js
Doesn't work, and it I'm not sure how well that idea is going to work with different CLI modes PM2 can launch the app. 
2. https://www.npmjs.com/package/windows-kill - native nodejs plugin (up to API 67?).
Can actually invoke the signal handler in the nodejs, but it does so for all child processes too, which makes it unusable in Cluster mode - signal will be received by all instances.
But that module replaces global `process.kill`, so I thought about doing it manually.
NOTE; If you play with the `process.kill()`, you have to disable `treekill`, because it uses Windows utility by default instead.
3. `process.kill()` substitution. You can wrap PM2 startup in another script and replace `process.kill` there. This will still allow running PM2 as a service, and not only in the standalone mode.
The idea there is to handle the `SIGINT` signal to notify an app that is about to be closed, and rely on the PM2 kill timeout to send the `SIGTERM` and explicitly close the app.
It's actually a 'reasonable' workaround, but you can't use any messaging methods (like `sendDataToProcessId` or `trigger` custom action), because the process is marked as `stopping` internally and all communication to such processes is blocked: https://github.com/Unitech/pm2/blob/91470b622a40f81091bbbaea4584f44ec16ee5b4/lib/God/ActionMethods.js#L714
So you eventually will have to get access to the `God.clusters_db[]` object to be able to use `process.send` - this is not 'trivial'.
4. Since the `process.kill()` replacement is quirky and is tightly coupled with PM2 version (relies on veriable names and etc.), you can notify the specific app instance with the `sendDataToProcessId` or another way, if you can get the instance that is going to be shutdown.
It's easy to guess, however, also PM2 version-specific, but somewhat less than previous item.
Let's say you want to stop/scale down an app. You need to call the `describe` or `list` first to get ID of the instances and their order. PM2 internally represents this list in the same order it kills apps - it uses filtered arrays and a simple `for` loop, so it's easy to guess and pre-notify apps. And you can rely on the 'shutdown' message for the auto-restarts.

---

I thought about fixing the higher-level abstractions, but it feels like that the change in that business logic is too risky and won't be accepted (considering the previous PRs I've seen):
`God.killProcess` is called only from the `God.stopProcessId`, so it could be refactored to accept switch to soft/hard stop the app. But if you do that, there is no need in the `softReload` function, and entire `startOrGracefulReload` feature relies on it.
And it's hard to say what kind of regressions could be introduced by affecting the timing of the `shutdown` message.

---

*Proposed solution*:

I think the easiest way to provide a Windows-specific workaround is to introduce a new global switch (`PM2_KILL_USE_SEND` environment variable, name TBD) to change how `God.killProcess` works - it's intent actually not to kill the process, but to notify it, and then the `God.processIsDead` method explicitly kills the process on timeout, if it's not dead.
So, by allowing `God.killProcess` to use `process.send()` instead of `process.kill()`, apps can be notified and then killed by timeout, if needed.
It's also possible to change the shutdown message with the existing `PM2_KILL_SIGNAL` variable.

Sample app:
```
setInterval(() => {}, 1000);

process.on('message', msg => {
  if (msg === 'SIGINT') {
    // Perform cleanup, app will be automatically terminated on kill timeout.
  }
});
```

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -100,6 +100,9 @@ var csts = {
 
   WORKER_INTERVAL         : process.env.PM2_WORKER_INTERVAL || 30000,
   KILL_TIMEOUT            : process.env.PM2_KILL_TIMEOUT || 1600,
+  KILL_SIGNAL             : process.env.PM2_KILL_SIGNAL || 'SIGINT',
+  KILL_USE_SEND           : process.env.PM2_KILL_USE_SEND || false,
+
   PM2_PROGRAMMATIC        : typeof(process.env.pm_id) !== 'undefined' || process.env.PM2_PROGRAMMATIC,
   PM2_LOG_DATE_FORMAT     : process.env.PM2_LOG_DATE_FORMAT !== undefined ? process.env.PM2_LOG_DATE_FORMAT : 'YYYY-MM-DDTHH:mm:ss'
 
--- a/lib/God/Methods.js
+++ b/lib/God/Methods.js
@@ -207,18 +207,32 @@ module.exports = function(God) {
   God.killProcess = function(pid, pm2_env, cb) {
     if (!pid) return cb({msg : 'no pid passed or null'});
 
-    var mode = pm2_env.exec_mode;
+    if (cst.KILL_USE_SEND && typeof(pm2_env.pm_id) === 'number') {
+      var proc = God.clusters_db[pm2_env.pm_id];
+
+      if (proc && proc.send) {
+        try {
+          proc.send(cst.KILL_SIGNAL);
+        } catch (e) {
+          console.error('[SimpleKill] %s pid can not be killed', pid, e.stack, e.message);
+        }
+        return God.processIsDead(pid, pm2_env, cb);
+      }
+      else {
+        console.log('[SimpleKill] %s pid cannot be notified with send()', pid);
+      }
+    }
 
     if (pm2_env.treekill !== true) {
       try {
-        process.kill(parseInt(pid), process.env.PM2_KILL_SIGNAL || 'SIGINT');
+        process.kill(parseInt(pid), cst.KILL_SIGNAL);
       } catch(e) {
         console.error('[SimpleKill] %s pid can not be killed', pid, e.stack, e.message);
       }
       return God.processIsDead(pid, pm2_env, cb);
     }
     else {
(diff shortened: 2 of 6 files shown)


Unitech:
thanks for this PR

PR #4596, opened 2020-02-03
Added filters for global envs

inerc:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| License       | MIT
| Doc PR        | https://github.com/pm2-hive/pm2-hive.github.io/pull/179


lib/Common.js
```diff
@@ -187,8 +187,17 @@ Common.prepareAppConf = function(opts, app) {
   else
     env = process.env;
 
-  // Change to double check  (dropped , {pm_cwd: cwd})
-  app.env = [{}, env, app.env || {}].reduce(function(e1, e2){
+  function filterEnv(envObj){
+    let new_env = {};
```
dipiash: In all code use `var` declaration. Why you use `let`?

inerc: fixed @dipiash

nlopin:
Is it going to be merged any time soon?

lib/binaries/CLI.js
```diff
@@ -42,6 +42,7 @@ commander.version(pkg.version)
   .option('--env <environment_name>', 'specify which set of environment variables from ecosystem file must be injected')
   .option('-a --update-env', 'force an update of the environment with restart/reload (-a <=> apply)')
   .option('-f --force', 'force actions')
+  .option('--filter_env [envs]', 'filter outgoing global values by prefix', function(v, m) { m.push(v); return m;}, [])
```
nlopin: We use `includes` in the `filterEnv` function and it checks if the string contains passed substring. So we filter outgoing values by the presence of the provided value in the key.

inerc: I hope that it will be in the release soon....

nlopin: @inerc I hope so, my pm2 crashes because of the env variables overflow. Can you change the description above? What about: `'filter out outgoing global values that contain provided strings'`?

inerc: Yes, it will be better. I have made changes.

Unitech:
thanks for the work.
I reworked the PR to fix some issues (allow boolean and string) and add some tests:
https://github.com/Unitech/pm2/commit/c93f0117865f48fe6dcfc5b2c3b8f29a8b246cd0

PR #4639, opened 2020-03-22
Fix --cwd CLI option description

warpdesign:
The `cwd` CLI option description appears to be the wrong one:

`run target script as <username>` -> `run target script from path <cwd>`

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT


Unitech:
good catch

Issue #4652, opened 2020-03-30
Allow any name for javascript ecosystem file.


AoDev:
## What's going wrong?
__Context__: I have a lot of troubles having pm2 work within a ES6 written codebase  (`import`). (experimental module flag, type: "module" in package.json, etc. I know there are other issues about it.

__I'd like__: to be able to use the `.cjs` extension for `ecosystem.config.cjs`
But it seems that the file is not found and it defaults to package.json where I have no config for pm2.

That said: per the docs:
> Note that using a Javascript configuration file requires to end the file name with .config.js

Not too sure why we can not use any name we want, but if this is hard coded, I guess that's the reason it's not working.

Can we do something about it please?

PS: pm2 is awesome ;)

Timic3:
Check out #4662. If you need it before (or if?) it gets merged, use: `npm install git://github.com/Timic3/pm2.git#fix/common-module-import -g`

AoDev:
@Timic3 Awesome. Thanks  : )

Timic3:
This has been merged into development branch, so you can now install it with `npm install git://github.com/Unitech/pm2.git#development -g`. This issue can now be closed. :)

PR #4662, opened 2020-04-12
Add .cjs (common JS) as a viable extension

Timic3:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | n/a
| Fixed tickets | #4652 
| License       | MIT

This pull request fixes the following problem: If you define `type: module` inside package.json, none of the extensions on config file would work (tried with mjs and js). This happens on `pm2 deploy` command.

This is merely a temporary solution; it would be nice to support module exports, but due to require being used [here](https://github.com/Unitech/pm2/blob/master/lib/Common.js#L289), we can't. I believe this happens only on `pm2 deploy` command.

### DIFF
--- a/lib/Common.js
+++ b/lib/Common.js
@@ -251,6 +251,8 @@ Common.isConfigFile = function (filename) {
     return 'yaml';
   if (filename.indexOf('.config.js') !== -1)
     return 'js';
+  if (filename.indexOf('.config.cjs') !== -1)
+    return 'js';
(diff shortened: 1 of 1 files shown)


koistya:
`pm2 deploy production` doesn't seem to be able find `ecosystem.config.cjs` in an ESM project.

PR #4593, opened 2020-02-02
fix for huntr

adunkey:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1234, #5678
| License       | MIT
| Doc PR        | https://github.com/pm2-hive/pm2-hive.github.io/pulls

PR #4767, opened 2020-07-02
fix: backward compatibility fix for tarball modules

ykolbin:
#4593 affected backward compatibility, if someone try to install tarball module without tarball option. 

Actually, it contains bad regular expression for following use cases:
1) Module name with digits (`pm install pm2-module-name.tar.gz`)
2) Modules/TAR.js can use Modularizer module_name as module_filepath, but #4593 regular expression is not allowed to specify full path to module (`pm install full/path/to/module/pm2-module-name.tar.gz`)
3) Modules/TAR.js can accept URL (`pm install http://example.com/pm2/pm2-module-name.tar.gz`)

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT

### DIFF
--- a/lib/API/Modules/Modularizer.js
+++ b/lib/API/Modules/Modularizer.js
@@ -39,7 +39,7 @@ Modularizer.install = function (CLI, module_name, opts, cb) {
     Common.logMod(`Installing local NPM module`);
     return NPM.localStart(CLI, opts, cb)
   }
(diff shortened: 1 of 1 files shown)


ykolbin:
Workaround: `pm install --tarball full/path/to/module/pm2-module-name.tar.gz`
@adunkey, do you remember original issue for your fix?

Issue #4781, opened 2020-07-15
Running reloadLogs while run as a non-root user causes process to not release and hang


mbrooks:
## What's going wrong?
Running reloadLogs while run as a non-root user causing process to not release and hang. 

## How could we reproduce this issue?
Reproducible in the FreePBX 15 Distro running the pm2 module version 15.0.3.7. Upgraded module manually to PM2 4.4.0 and the issue still persists.

## Supporting information
Logs:
```
2020-07-14T20:57:49: PM2 log: App [ucp:1] starting in -fork mode-
2020-07-14T20:57:50: PM2 log: App [core-fastagi:0] online
2020-07-14T20:57:50: PM2 log: App [ucp:1] online
2020-07-14T20:57:50: PM2 log: App [xmpp:2] starting in -fork mode-
2020-07-14T20:57:50: PM2 log: App [xmpp:2] online
2020-07-14T20:58:29: PM2 log: Reloading logs...
2020-07-14T20:58:29: PM2 log: Reloading logs for process id 0
2020-07-14T20:58:29: PM2 log: Reloading logs for process id 1
2020-07-14T20:58:29: PM2 log: Reloading logs for process id 2
2020-07-14T20:58:32: PM2 log: Reloading logs...
2020-07-14T20:58:32: PM2 log: Reloading logs for process id 0
2020-07-14T20:58:32: PM2 log: ===============================================================================
2020-07-14T20:58:32: PM2 log: --- PM2 global error caught ---------------------------------------------------
2020-07-14T20:58:32: PM2 log: Time                 : Tue Jul 14 2020 20:58:32 GMT+0000 (UTC)
2020-07-14T20:58:32: PM2 error: Cannot read property 'destroy' of undefined
2020-07-14T20:58:32: PM2 error: TypeError: Cannot read property 'destroy' of undefined
    at ChildProcess.cspr._reloadLogs (/var/www/html/admin/modules/pm2/node/node_modules/pm2/lib/God/ForkMode.js:288:23)
    at /var/www/html/admin/modules/pm2/node/node_modules/pm2/lib/God/ActionMethods.js:651:19
    at Array.forEach (<anonymous>)
    at God.reloadLogs (/var/www/html/admin/modules/pm2/node/node_modules/pm2/lib/God/ActionMethods.js:633:16)
    at Server.onmessage (/var/www/html/admin/modules/pm2/node/node_modules/pm2-axon-rpc/lib/server.js:104:6)
[...]

PR #4782, opened 2020-07-15
Prevent reloadLogs command from hanging

mbrooks:
Previously, when the pm2 reloadLogs command was running as a non-root
user, the process will not release and hangs. The reason for the hang
is that the forked process never closes because the object is empty
and doesn't contain the expected object. This commit checks if an
object exists first, before trying to execute methods on it.

Related to: #4781

Co-authored-by: Franck Danard <fdanard@sangoma.com>

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | #4781
| License       | MIT

### DIFF
--- a/lib/God/ForkMode.js
+++ b/lib/God/ForkMode.js
@@ -275,21 +275,23 @@ module.exports = function ForkMode(God) {
       cspr.once('exit', function forkClose(status) {
         try {
           for(var k in stds){
-            if (stds[k].destroy) stds[k].destroy();
-            else if (stds[k].end) stds[k].end();
-            else if (stds[k].close) stds[k].close();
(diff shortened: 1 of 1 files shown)


mbrooks:
Hello, Does the CLA need to be signed by me or the responsible party in my company?

mbrooks:
If there are no more issues with the CLA, maybe someone can help me on the right direction for this issue? This patch seems to resolve the issue for us, but it may not be the best approach. If you can point me in a direction I should look, it would be helpful. I would love to help contribute if I can to resolve our issue.

Thanks!

Unitech:
excellent, thank you

PR #4889, opened 2020-10-25
Fix: filter_env not work when use {filter_env:true}

bugkiwi:
filter_env not work when use {filter_env:true}; {filter_env:somestring} should not change process.env

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | no
| License       | MIT

### DIFF
--- a/lib/Common.js
+++ b/lib/Common.js
@@ -193,20 +193,20 @@ Common.prepareAppConf = function(opts, app) {
     if (app.filter_env == true)
       return {}
 
-    if (typeof app.filter_env === 'string') {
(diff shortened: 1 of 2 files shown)


mhassan1:
I was also bitten by this bug. It might be a good idea to open an issue for tracking purposes.

By the way, I noticed that `filter_env: ['']` got me the expected behavior of `filter_env: true`. It's ugly, but it works.

PR #5141, opened 2021-08-03
fix(lib): write pid file

petrgrishin:
pm2@4.2.1 and nodejs v14+


orgads:
Fixed in e41282df4cd3a76c8950258a0ede2e6d79d81414. Can be closed.

PR #5150, opened 2021-08-13
Fix typo

dsiddy:
Not much to say, impact-wise, just a quick update to the error output to fix a typo that I might easily have made myself. 😊


dsiddy:
I'm curious as to why merely changing a couple of constant strings would cause the `should automatically activate source map support (detect main.ts)` test to fail. Pretty sure it's a false negative.

PR #5307, opened 2022-02-17
fix: TypeScript types for error callbacks

sheepsteak:
Previously opened as #4882 but it got closed when the `development` branch was deleted.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | maybe
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT

`Error` will not always exist and can be `null`.

With regards to BC, anyone using TypeScript and PM2 would already be using something like `if(err) {…}` so it should be fine.


sheepsteak:
Closing in favour of #5350.

PR #5350, opened 2022-04-16
Update Typings for Custom Instances of PM2

sasial-dev:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | Supersedes #4882, #5307
| License       | MIT

Allows TS to have pm2.custom typed and supersedes #4882, #5307.


types/API.d.ts
```diff
@@ -0,0 +1,487 @@
...
+   */
+  updateEnv?: boolean;
+}
+
+// Types
+
+type ProcessStatus = 'online' | 'stopping' | 'stopped' | 'launching' | 'errored' | 'one-launch-status';
+type Platform = 'ubuntu' | 'centos' | 'redhat' | 'gentoo' | 'systemd' | 'darwin' | 'amazon';
+
+type ErrCallback = (err: Error) => void;
```
sheepsteak, line 482: Don't you need `Error | null` here too? `connect` uses this and won't always return an error.

sasial-dev: Out of scope for this PR, as that's the already existing typings iirc

sheepsteak: It is the existing type but you've changed the ones below anyway so may as well do this one.

Existing typings:
https://github.com/Unitech/pm2/blob/2573516e9321a78fb10474ea58c2cb487a663de6/types/index.d.ts#L448-L453

hinogi (approved):
LGTM

PR #5520, opened 2023-01-11
Autho flush stdout w/bash

vitomadio:
This change solves the problem, python does not print/stdout on pm2 logs when the project is started with the bash interpreter


hinogi (approved):
LGTM

vitomadio:
Is this PR gonna be merged some day?

Issue #5529, opened 2023-01-18
pm2 flush doesn't work as expected


Sailboat265:
## What's going wrong?
`pm2 flush <app_id>` or `pm2 flush <app_name>` doesn't appear to be flushing the logs successfully.

Linked issues: #4074, #4553, #4696
## How could we reproduce this issue?
At pm2 version of 5.2.2, just try to run either `pm2 flush <app_id>` or `pm2 flush <app_name>`.

After that, do `pm2 log <app_id>` or `pm2 log <app_name>`, you will see the logs aren't flushed and still remained untouched.
## Supporting information

```
--- PM2 report ----------------------------------------------------------------
Date                 : Thu Jan 19 2023 07:04:33 GMT+0800 (Malaysia Time)
===============================================================================
--- Daemon -------------------------------------------------
pm2d version         : 5.2.2
node version         : 17.9.0
node path            : not found
argv                 : /usr/bin/node,/usr/local/share/.config/yarn/global/node_modules/pm2/lib/Daemon.js
argv0                : node
user                 : pi
uid                  : 1000
gid                  : 100
uptime               : 2771min
===============================================================================
--- CLI ----------------------------------------------------
local pm2            : 5.2.2
node version         : 17.9.0
node path            : /usr/local/bin/pm2
argv                 : /usr/bin/node,/usr/local/bin/pm2,report
argv0                : node
user                 : pi
uid                  : 1000
gid                  : 100
===============================================================================
--- System info --------------------------------------------
arch                 : arm64
platform             : linux
type                 : Linux
cpus                 : Cortex-A72
cpus nb              : 4
freemem              : 3414192128
totalmem             : 3977629696
home                 : /home/pi
===============================================================================
--- PM2 list -----------------------------------------------
┌─────┬──────────────────┬─────────────┬─────────┬─────────┬──────────┬────────┬──────┬───────────┬──────────┬──────────┬──────────┬──────────┐
│ id  │ name             │ namespace   │ version │ mode    │ pid      │ uptime │ ↺    │ status    │ cpu      │ mem      │ user     │ watching │
├─────┼──────────────────┼─────────────┼─────────┼─────────┼──────────┼────────┼──────┼───────────┼──────────┼──────────┼──────────┼──────────┤
│ 0   │ Api-Alpha        │ default     │ 1.0.0   │ fork    │ N/A      │ 0      │ 0    │ stopped   │ 0%       │ 0b       │ pi       │ disabled │
[...]

PR #5530, opened 2023-01-18
Fixes pm2 not flushing logs

Sailboat265:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | not ran
| Fixed tickets | #5529
| License       | MIT
| Doc PR        | N/A

<h3>What's this PR about?</h3>

This PR aims to fix and restore the ability to use either `pm2 flush <app_id>` or `pm2 flush <app_name>` to flush logs successfully.

<h3>Why so many commits?</h3>

In the first commit of this PR, I used a regex to extract a valid filename from a log path to allow the existing `if-else` comparison to continue working whatever it's intended.

It worked perfectly. Well only if the user didn't seek to customize their logging path or name! Because it's found out later that this solution isn't feasible as it would mean that a log file *must* end with an extension of `.log`, or else the flush will not work.

An example is when a user tries to redirect stderr to a file like `~/myApp/logs/process.error`, the solution will not be able to handle this. Hence, I have chosen to revert the changes made by the first commit.  

<h3>About the latest commit</h3>

Now, about the latest commit of this PR. The printout msg of flushing a log file has been moved inside their respective `if` block. This fixed a problem where users are reporting that the command is printing out the log file message like below even though the logs aren't even getting flushed.

```
[PM2] Flushing:
[PM2] /home/pi/.pm2/logs/My-App-out.log
[PM2] /home/pi/.pm2/logs/My-App-error.log
[PM2] Logs flushed
```

And I don't really get what's this `if-else` trying to accomplish within its logic comparison.

``` javascript
if (l.pm2_env.pm_out_log_path.lastIndexOf('/') < l.pm2_env.pm_out_log_path.lastIndexOf(api)) {
    <...>
}
```

Anyways, we will be replacing it and use `fs.existsSync(FILE_PATH)` to check whether a log file exists or not. So now, part of the code looks like this:
``` javascript
if (fs.existsSync(l.pm2_env.pm_out_log_path)) {
    Common.printOut(cst.PREFIX_MSG + l.pm2_env.pm_out_log_path);
    fs.closeSync(fs.openSync(l.pm2_env.pm_out_log_path, 'w'));
}
```

This solution should be universal enough that it allows the command to flush the logs correctly even though a unique log path has been specified.

### DIFF
--- a/lib/API/LogManagement.js
+++ b/lib/API/LogManagement.js
@@ -41,21 +41,23 @@ module.exports = function(CLI) {
           fs.closeSync(fs.openSync(l.pm2_env.pm_out_log_path, 'w'));
           fs.closeSync(fs.openSync(l.pm2_env.pm_err_log_path, 'w'));
         }
-        else if (l.pm2_env.name === api) {
+        else if (l.pm2_env.pm_id == api || l.pm2_env.name === api) {
           Common.printOut(cst.PREFIX_MSG + 'Flushing:');
-          Common.printOut(cst.PREFIX_MSG + l.pm2_env.pm_out_log_path);
-          Common.printOut(cst.PREFIX_MSG + l.pm2_env.pm_err_log_path);
 
-          if (l.pm2_env.pm_log_path &&
-              l.pm2_env.pm_log_path.lastIndexOf('/') < l.pm2_env.pm_log_path.lastIndexOf(api)) {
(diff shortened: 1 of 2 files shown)


Sailboat265:
This pull request is too polluted and it seems like I didn't adhere to the CONTRIBUTING rules too.

Hence, I have decided to move to a new PR, #5533.

Issue #5788, opened 2024-03-18
`pm2 logs` with a namespace doesn't work if run before relevant `pm2 start`


bawjensen:
## What's going wrong?

If you try to start the logging before you've started the processes (perhaps because you want to make sure you don't miss any logs), then it breaks when provided with a namespace.

## How could we reproduce this issue?

Should be simple enough to reproduce, e.g. `pm2 logs my-namespace` and `pm2 start index.js --namespace my-namespace`.

## Supporting information

```
# Run the following commands
$ pm2 report
```

```
--- PM2 report ----------------------------------------------------------------
Date                 : Mon Mar 18 2024 15:29:19 GMT-0700 (Pacific Daylight Time)
===============================================================================
--- Daemon --------------------------------------------------------------------
pm2d version         : 5.3.1
node version         : 20.8.0
node path            : /Users/bryan/src/coda/build/node/bin/node
argv                 : /Users/bryan/src/coda/build/node-v20.8.0-darwin-arm64/bin/node,/Users/bryan/sandbox/pm2/lib/Daemon.js
argv0                : node
user                 : bryan
uid                  : 503
gid                  : 20
uptime               : 13min
===============================================================================
--- CLI -----------------------------------------------------------------------
local pm2            : 5.3.1
node version         : 20.8.0
node path            : not found
argv                 : /Users/bryan/src/coda/build/node-v20.8.0-darwin-arm64/bin/node,/Users/bryan/src/coda/node_modules/pm2/bin/pm2,report
argv0                : node
user                 : bryan
uid                  : 503
gid                  : 20
===============================================================================
--- System info ---------------------------------------------------------------
arch                 : arm64
platform             : darwin
type                 : Darwin
cpus                 : Apple M1 Max
cpus nb              : 10
freemem              : 7786971136
totalmem             : 68719476736
home                 : /Users/bryan
===============================================================================
--- PM2 list ------------------------------------------------------------------
┌────┬───────────┬─────────────┬─────────┬─────────┬──────────┬────────┬──────┬───────────┬──────────┬──────────┬──────────┬──────────┐
│ id │ name      │ namespace   │ version │ mode    │ pid      │ uptime │ ↺    │ status    │ cpu      │ mem      │ user     │ watching │
└────┴───────────┴─────────────┴─────────┴─────────┴──────────┴────────┴──────┴───────────┴──────────┴──────────┴──────────┴──────────┘
===============================================================================
--- Daemon logs ---------------------------------------------------------------
[...]

PR #5660, opened 2023-08-19
Fix matching logic for logs from namespace when lines = 0

bawjensen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #5658
| License       | MIT
| Doc PR        | N/A (?)

PR #5789, opened 2024-03-18
Fix bug with namespace when logs command run before start

bawjensen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #5788
| License       | MIT
| Doc PR        | 

I created the ticket linked above, and this is my attempt at a fix.

This implementation is something my company has been using (via `pnpm patch`) for a few months now to good effect. Others might benefit from it too.

I tried to do something similar to what I did in my past PR (#5660), since that one seemed rather painless.

### DIFF
--- a/lib/API/LogManagement.js
+++ b/lib/API/LogManagement.js
@@ -161,7 +161,6 @@ module.exports = function(CLI) {
     // Get the list of all running apps
     that.Client.executeRemote('getMonitorData', {}, function(err, list) {
       var regexList = [];
-      var namespaceList = [];
 
       if (err) {
         Common.printError(err);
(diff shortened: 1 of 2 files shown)


bawjensen:
> I tried to do something similar to what I did in my past PR (#5660), since that one seemed rather painless.

perhaps worth noting: this fix works by deleting code. That's only possible because of the specifics of the change in #5660, i.e. my last pm2 contribution. I opted to split them up, and this second half took me honestly too long to create.

bawjensen:
@Unitech checking in again on this PR. Let me know if there's anything I can do

Issue #5618, opened 2023-06-07
How to setup timezone for auto restart?


trandinhthangv2:
How to setup timezone for auto restart?

```json
{
  apps: [
    {
      script: 'index.js',
      cron_restart: '30 22 * * *',
      env: {
        TZ: 'America/Los_Angeles',
      },
    },
  ],
};
```
I write code as above but not working

KamranBiglari:
@trandinhthangv2 Did you find a way to set the timezone for cron_restart ?

Vincent-SV:
Even documenting which time is used (I think it is the system time ?) would be a great improvement, I couldn't find anything on https://pm2.keymetrics.io/docs/usage/restart-strategies/

mgiova:
I needed this option as well, so I implemented it in PR https://github.com/Unitech/pm2/pull/5877

fwo88:
up!

PR #5877, opened 2024-07-31
Add app cron_timezone option to change the timezone used in cron_restart

mgiova:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #5618
| License       | MIT
| Doc PR        | 

Sometimes you need to restart a process on a time based on a different time-zone than the one used by the server and DST differences could cause problems. This PR adds an option to specify that time-zone that is passed as-is to the underlying Croner lib.


gkourpa:
This is a very basic feature and should definitely be added.

PR #6064, opened 2025-12-03
Update type definitions to include `custom` export

fredrikekelund:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

> [!TIP]
> Exclude whitespace changes when looking at the PR diff to see that I haven't actually changed all that much. 

pm2 [exports](https://github.com/Unitech/pm2/blob/ff1ca974afada8730aa55f8ed1df40e700cedbcb/index.js) two members in the programmatic interface: `default` (an instance of `API`) and `custom` (the `API` class). However, the `custom` prop is not included in `types/index.d.ts`. 

This PR fixes that problem by moving all function definitions inside an `API` class in `types/index.d.ts`. We then use this single definition to declare both the `default` and `custom` exports. 

I also took the opportunity to update the `start` type definitions to take an `ErrProcsCallback` as the second parameter (a callback where the second argument is `Proc[]` instead of `Proc`).

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -3,315 +3,387 @@
 
 // Exported Methods
 
-/**
- * Either connects to a running pm2 daemon (“God”) or launches and daemonizes one.
- * Once launched, the pm2 process will keep running after the script exits.
- * @param errback - Called when finished connecting to or launching the pm2 daemon process.
- */
-export function connect(errback: ErrCallback): void;
-/**
- * Either connects to a running pm2 daemon (“God”) or launches and daemonizes one.
- * Once launched, the pm2 process will keep running after the script exits.
- * @param noDaemonMode - (Default: false) If true is passed for the first argument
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -711,12 +783,33 @@ export interface InstallOptions {
...
+  | "one-launch-status"
+  | "waiting_restart";
+type Platform =
+  | "ubuntu"
+  | "centos"
+  | "redhat"
+  | "gentoo"
+  | "systemd"
+  | "darwin"
+  | "amazon";
```
fredrikekelund, line 801: Several parts of this diff are just auto-formatting. This is one example.

Issue #6014, opened 2025-07-14
support node run typescript file


JYbill:
- node v24 can run ts file without ts-node/tsx...
- pm2 using node exec ts file will get error, It does not affect the startup of the cluster
```
2025-07-14T09:13:48: PM2 log: App [wzj2:4] starting in -cluster mode-
Failed to load Typescript interpreter: Cannot find module 'ts-node/register'
Require stack:
- /usr/local/lib/node_modules/pm2/lib/ProcessContainer.js

# in docker container
pm2 list
┌────┬──────────────────┬─────────────┬─────────┬─────────┬──────────┬────────┬──────┬───────────┬──────────┬──────────┬──────────┬──────────┐
│ id │ name             │ namespace   │ version │ mode    │ pid      │ uptime │ ↺    │ status    │ cpu      │ mem      │ user     │ watching │
├────┼��─────────────────┼─────────────┼─────────┼─────────┼──────────┼────────┼──────┼───────────┼──────────┼──────────┼──────────┼──────────┤
│ 1  │ wzj2             │ default     │ 2.0.0   │ cluster │ 188      │ 3m     │ 1    │ online    │ 0%       │ 254.5mb  │ root     │ disabled │
│ 2  │ wzj2             │ default     │ 2.0.0   │ cluster │ 189      │ 3m     │ 1    │ online    │ 0%       │ 179.1mb  │ root     │ disabled │
│ 3  │ wzj2             │ default     │ 2.0.0   │ cluster │ 204      │ 3m     │ 1    │ online    │ 0%       │ 258.0mb  │ root     │ disabled │
│ 4  │ wzj2             │ default     │ 2.0.0   │ cluster │ 205      │ 3m     │ 1    │ online    │ 0%       │ 280.9mb  │ root     │ disabled │
└────┴──────────────────┴─────────────┴─────────┴─────────┴──────────┴────────┴──────┴───────────┴──────────┴──────────┴──────────┴──────────┘

```
> I didn't see any instructions in the document regarding the default use of the ts loader, its very strange. 
> maybe we cloud use node.js exec typescript file.

PR #6083, opened 2026-02-16
Fixed ridiculous default of using bun to run .ts files and removed ts-node/register entirely

mdierolf:
PM2 has been broken for nodejs/typescript users for at least a year, this fixes it. https://github.com/Unitech/pm2/issues/6014

ts-node/register? gone. I don't want it. Node runs .ts files just fine. If it is needed put it behind some kind of config option.
bun? removed as the default for .ts files because this was undocumented behavior and frankly a rather ridiculous choice to force upon users.

### DIFF
--- a/lib/API/Modules/LOCAL.js
+++ b/lib/API/Modules/LOCAL.js
@@ -20,7 +20,7 @@ var INTERNAL_MODULES = {
   'event-loop-inspector': {name: 'event-loop-inspector'},
   'v8-profiler': {name: 'v8-profiler-node8'},
   'profiler': {name: 'v8-profiler-node8'},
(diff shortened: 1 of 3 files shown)


niksy:
What is the status of this PR? This really shouldn’t be default or at least should be configurable. Node handles TS perfectly fine in latest versions.

EDIT: from what I can see, this has been resolved in 7.0.4: https://github.com/Unitech/pm2/commit/1dbdea617046ef67a7f1409be4ad84b896f2dfd0

Unitech:
@niksy so all is fine on your side ?
If yes I will close that PR 
Thanks for reporting @mdierolf

PR #6121, opened 2026-06-22
Fix --require scripts missing env vars in cluster mode

tlhunter:
This allows scripts that run via `--require` to access the intended ENV vars supplied by PM2.

We had a user-submitted issue that the Datadog tracer wouldn't properly read ENV var configuration when used with PM2. While [we can fix this by having Datadog do the pm2_env parsing](https://github.com/DataDog/dd-trace-js/pull/8863), it does seem like similar functionality would need to land in other libraries as well. For that reason it seems better to try and solve it inside of pm2.

### DIFF
--- a/lib/God/ClusterMode.js
+++ b/lib/God/ClusterMode.js
@@ -11,6 +11,7 @@
  * @project PM2
  */
 var cluster       = require('cluster');
+var path          = require('path');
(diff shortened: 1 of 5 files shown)


tlhunter:
@Unitech Any chance of getting a review on this?

PR #6136, opened 2026-07-23
fix(cli): add start-or-reload kebab-case alias

### LINKED ISSUE #6107: CLI naming inconsistency: startOrReload should be start-or-reload
## Summary

The `startOrReload` command uses camelCase, which is inconsistent with every other PM2 CLI command (`start`, `stop`, `restart`, `reload`, `delete`, etc.) and violates the universal convention for CLI argument naming.

## The problem

CLI arguments live in shell scripts, documentation, and muscle memory. camelCase is hostile in that environment:

- You can never remember the exact casing without checking the docs (`startOrReload`? `startorreload`? `start-or-reload`?)
- It breaks the reasonable assumption that CLI args are case-insensitive
- Tab completion behaves unexpectedly
- It's a clear sign that the command was added by a different contributor without a CLI style guide enforcing consistency

## Suggested fix

Rename to `start-or-reload` (kebab-case), which is the universal CLI convention, while keeping `startOrReload` as a deprecated alias for backwards compatibility.

```bash
# Current (hostile)
pm2 startOrReload ecosystem.config.js --only my-service

# Proposed (consistent with every other pm2 command)
pm2 start-or-reload ecosystem.config.js --only my-service
```

## Note

This issue was filed by an AI assistant (Claude) on behalf of @johanatan, who is off enjoying life somewhere and asked me to flag this. The technical observation is his; I'm just the one typing.

Hashim1999164:
## Summary
- Prefer `start-or-reload` as the CLI command name so it matches other kebab-style PM2 commands.
- Keep `startOrReload` as an alias so existing scripts and docs keep working.

Fixes #6107

## Test plan
- [ ] `pm2 start-or-reload ecosystem.config.js` runs the same path as before
- [ ] `pm2 startOrReload ecosystem.config.js` still works via the alias
- [ ] `pm2 --help` lists `start-or-reload`

### DIFF
--- a/lib/binaries/CLI.js
+++ b/lib/binaries/CLI.js
@@ -328,7 +328,8 @@ commander.command('startOrRestart <json>')
     pm2._startJson(file, commander, 'restartProcessId');
   });
 
-commander.command('startOrReload <json>')
+commander.command('start-or-reload <json>')
+  .alias('startOrReload')
   .description('start or gracefully reload JSON file')
   .action(function(file) {
     pm2._startJson(file, commander, 'reloadProcessId');


goransh-walia:
This looks good to me at a glance — thanks for putting it together.

PR #6143, opened 2026-08-08
Fix two ReferenceErrors in error paths

isamu:
Both of these sit in code that only runs after something else has already failed, so instead of the message the author wrote, the user gets a stack trace. I could not find an existing issue for either.

Found by running ESLint's `no-undef` over the tree.

---

### 1. `lib/API/Extra.js:703` — `conf` is not defined

```js
that.exitCli(conf.ERROR_EXIT);
```

Every other exit in this file uses `cst.ERROR_EXIT`, and `cst` is what the file requires (`var cst = require('../../constants.js')` on line 8). `conf` is not defined anywhere in it.

Reached when `getMonitorData` fails during `pm2 monit`: the error is logged, and then the process throws `ReferenceError: conf is not defined` instead of exiting with `ERROR_EXIT`.

### 2. `lib/completion.js:203` — `completer` is not defined

`writeRc` looks like a copy of `readRc` with one argument removed, but the body still reads it:

```js
function readRc(completer, cb) {                 // takes completer
  ...
  if (err) return cb(new Error("No " + file + " ... " + completer + " ..."));

function writeRc(content, cb) {                  // does not take completer
  ...
  if (err) return cb(new Error("No " + file + " ... " + completer + " ..."));
```

Reached from `install()` and `uninstall()` when `lstat` on the rc file fails — for example if `$HOME` has no rc file for the current `$SHELL`, or the file is removed between the read and the write.

Both call sites already have `completer` in scope, so threading it through is the smallest change:

```
$ node repro.js
  before: ReferenceError: completer is not defined
  after:  Error: No .zshrc file. You'll have to run instead: pm2 completion >> ~/.zshrc
```

---

Four lines in total. Happy to reshape it — the two functions could also share one stat-and-error helper so the copy cannot drift again, but I kept this minimal on purpose.

### DIFF
--- a/lib/API/Extra.js
+++ b/lib/API/Extra.js
@@ -700,7 +700,7 @@ module.exports = function(CLI) {
       that.Client.executeRemote('getMonitorData', {}, function(err, list) {
         if (err) {
           console.error('Error retrieving process list: ' + err);
-          that.exitCli(conf.ERROR_EXIT);
+          that.exitCli(cst.ERROR_EXIT);
         }
 
         Monit.refresh(list);
--- a/lib/completion.js
+++ b/lib/completion.js
@@ -160,7 +160,7 @@ function install(name, completer, cb) {
   function next() {
     if(!rc || !scriptOutput) return;
 
-    writeRc(rc + scriptOutput, function(err) {
+    writeRc(completer, rc + scriptOutput, function(err) {
       if(err) return cb(err);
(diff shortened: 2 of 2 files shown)


lbesecker195 (approved):
Verified both on current master: `lib/API/Extra.js` requires constants as `cst` (line 8) and has no `conf` binding, and `writeRc` in `lib/completion.js` references `completer` in its error path without it being in scope. Threading `completer` through matches how `readRc` already works. LGTM.

PR #6145, opened 2026-08-18
fix: clear watcher registry in disableAll

zloglevel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

## Summary

`God.watch._watchers` is initialized as an object keyed by `pm_id`, but
`God.watch.disableAll()` previously attempted to remove entries with
`Array.prototype.splice()`.

When the watcher registry is non-empty, the first watcher is closed and the method then throws:

```text
TypeError: watchers.splice is not a function
```

This prevents the remaining watchers from being closed and removed.

## Changes

- Replace the invalid `splice()` call with `Object.keys()` iteration and
  property deletion.
- Add a regression test with two fake watchers.
- Verify that each watcher is closed exactly once.
- Verify that the watcher registry is empty after `disableAll()` completes.

## Testing

```text
npx mocha --exit --bail test/interface/watcher.spec.mocha.js
```

Result: 1 passing.


zloglevel:
@Unitech Could you please take a look when you have a chance? Thanks!

zloglevel:
@Unitech  Hi, Could you please review this PR at your convenience? Thank you very much.

